PR #3, opened 2022-07-20
Update setup.sh

StevenDufresne:
- I'm not certain the data import was supposed to be commented out. 
- I also didn't have the front page ready to go on the first load.
- `wp option update page_on_front 8891` is not necessarily future-proof but works well enough for now.


ryelle (approved):
> I'm not certain the data import was supposed to be commented out.

🤦🏻 

Thanks for catching these

Issue #10, opened 2022-07-20
Create a shortcode for `WP_MARKET_SHARE`


ryelle:
This is a single defined constant that provides the current WP market share (ex, 43%). In `wporg-main`, this is used to build strings, so that they don't need to be re-translated every time the percentage changes. It also keeps the value defined in one place, so we don't end up with 40% in one place, and 43% in another.

```php
printf(
	/* translators: WordPress market share: 39 - Note: The following percent sign is '%%' for escaping purposes; */
	esc_html__( 'Use the software that powers over %s%% of the web.', 'wporg' ),
	esc_html( number_format_i18n( WP_MARKET_SHARE ) )
);
```

How should we handle this for content entered in the block editor?

tellyworth:
Could we do this with a few super-simple server side blocks? That basically just run your code snippet, or similar to it.

dd32:
IMHO; for the initial release I would just use an inline shortcode.

It looks like "tokens" will be shortcodes-v2 which would likely be the proper solution in future. https://github.com/WordPress/gutenberg/pull/42015

ryelle:
> IMHO; for the initial release I would just use an inline shortcode.

I agree. If we want to make it easier for editors we could add a toolbar button/dropdown like #16 to insert these, but spending too much time here doesn't make much sense if a core solution is on the way.

ryelle:
Renamed & moved this to "to do" since it sounds like we have a path forward.

PR #16, opened 2022-07-21
Try using a format for a market share placeholder

ryelle:
See #10 

I started trying something out and landed here. This uses the [Formatting API](https://developer.wordpress.org/block-editor/how-to-guides/format-api/) to insert an inline element we can control in JS.

I put [more of a write-up in the file](https://github.com/WordPress/wporg-main-2022/blob/try/formats-for-inline-values/source/wp-content/themes/wporg-main-2022/js/custom-formats.js), but to keep everything in one place I'll copy-paste it here.

This creates a button in the toolbar which inserts a custom, non-editable HTML element `<wporg-percentage />`, with the visible content of "43%". That could be a constant we pull from somewhere, but the string itself is saved in the content as text, and wouldn't live-update. For that, it would need to be replaced on the frontend too (a naive string replacement of the HTML element would probably be fine).

It renders something like this:

Screen Shot 2022-07-21 at 11 07 57 AM

I don't think this code should be used as-is, but the concept is a starting point for the issue in #10.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -6,6 +6,7 @@ namespace WordPressdotorg\Theme\Main_2022;
  * Actions and filters.
  */
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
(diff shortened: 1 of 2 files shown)


ryelle:
Closing this for now, though we could so something with it in the future.

Issue #1, opened 2022-07-20
Syncing content between local dev, staging, and production


tellyworth:
A key assumption for this project is that some of the design will be implemented in WordPress content, rather than in code (like Kelly's front-page-in-the-editor experiment).

We'll need a way to sync and share this kind of content in several ways:

- Between developers who are working on separate local dev environments
- From this repo to a staging site on w.org, prior to going live
- From the staging site to live production (maybe, depending on how we do this)

This issue is for discussing sync methods and workflow.

tellyworth:
Some possibilities for the first part, syncing content during early development, not exhaustive:

* Commit and edit content in text files, and have a script to import them into post content as part of the build process.
* Just start by putting everything in theme template files, and move parts of it into post content later.

ryelle:
I was thinking about something like this for hacky translation support, but it would also help keep the content synced… What if we had a set of PHP files that contain the block content + i18n functions, like this:

```php
<!-- wp:group {"style":{"spacing":{"padding":{"top":"6rem","right":"4rem","bottom":"6rem","left":"4rem"}}},"backgroundColor":"blue-5","textColor":"white","layout":{"inherit":true}} -->
<div
	class="wp-block-group has-white-color has-blue-5-background-color has-text-color has-background"
	style="padding-top: 6rem; padding-right: 4rem; padding-bottom: 6rem; padding-left: 4rem"
>
	<!-- wp:heading {"textAlign":"center","level":1} -->
	<h1 class="has-text-align-center"><?php _e( 'Meet WordPress', 'wporg' ); ?></h1>
	<!-- /wp:heading -->

	<!-- wp:paragraph {"align":"center","fontSize":"extra-large"} -->
	<p class="has-text-align-center has-extra-large-font-size">
		<?php _e( 'WordPress is open source software you can use to create a beautiful website, blog, or app.', 'wporg' ); ?>
	</p>
	<!-- /wp:paragraph -->
</div>
<!-- /wp:group -->
```

Could we use something like the filter approach used in [pattern translations](https://github.com/WordPress/pattern-directory/blob/6b7601f39815e3071a42ac1d6683e9e56eb0b773/public_html/wp-content/plugins/pattern-translations/pattern-translations.php#L97-L114) to swap the post content for our PHP content?

It's pretty much what's done in the theme now, but as block content instead of plain HTML.

------

[...]

PR #18, opened 2022-07-26
Add setup script and filter to import content from remote REST API

tellyworth:
This will import content directly from the staging server REST API.

See #1 

Depends on https://github.com/WordPress/wporg-mu-plugins/pull/227

That PR will need to be merged and synced to the server in order for the import script here to work.

### DIFF
--- a/env/import-content.php
+++ b/env/import-content.php
@@ -1,0 +1,94 @@
+#!/usr/bin/php
+<?php
+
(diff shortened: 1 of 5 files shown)


tellyworth:
To run it:

`yarn run setup:refresh`

Note that the server side mu-plugin has to be there for this to work. I'm not sure it's possible to test this with a sandbox.

If this works ok we could replace the data.xml import with it during initial setup.

PR #22, opened 2022-07-27
Tools: Merge existing content in when syncing from live site.

ryelle:
See #18. The sync script pulls down content from a given URL (the staging site, currently). This created duplicate content when run if you've already got content from `yarn setup:wp`. The IDs from the import into an empty site are the same, so we can just quickly use `get_post` to check for existing content. If we have an existing post, this merges the new content into the existing array, so that when `wp_insert_post` is run, it has the existing ID and updates it.

0. Clean your site with `yarn wp-env clean all`
1. Run `yarn setup:wp` to set up content
2. Run `yarn setup:refresh` to sync down the content -- you should now have a page called "Steve's test", but no duplicate pages.

### DIFF
--- a/env/import-content.php
+++ b/env/import-content.php
@@ -1,7 +1,8 @@
 #!/usr/bin/php
 <?php
+// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
 
-namespace WordPress_org\Parent_2021\ImportTestContent;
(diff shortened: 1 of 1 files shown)


tellyworth:
Looks good to me. I did wonder if we should handle conflicts (where local edits have been made) but that can be a separate issue if it's needed.

PR #25, opened 2022-07-28
Force the import script to use a production server

tellyworth:
This fixes errors when `wordpress.org` is locally sandboxed, since the sandbox server will reject un-proxied requests.

If you've sandboxed `wordpress.org` in /etc/hosts you'll get 403 errors when running `yarn setup:refresh`. That's because Docker doesn't use the proxy, so the sandbox rejects those REST API requests.

The PR uses CURLOPT_CONNECT_TO to bypass the sandbox and connect to a production server.

See #18, #22.

1. Sandbox `wordpress.org` in `/etc/hosts`
2. `yarn setup:refresh`

### DIFF
--- a/env/import-content.php
+++ b/env/import-content.php
@@ -50,7 +50,7 @@ function sanitize_meta_input( $meta ) {
  * Filter CURL requests to bypass sandboxes and always hit a production server.
  * Docker doesn't use the proxy, so those requests will fail when wordpress.org is sandboxed.
  */
-function filter_curl_options( &$ch ) {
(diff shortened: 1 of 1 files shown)


dd32 (approved):
As mentioned, while `wp_remote_get` might use a non-curl transport, this seems safe to me considering it's running within a Docker container.

I'm not sure it needs to be explicitly set to accept a reference, `&$ch`, as all objects (and therefor, Resources?) are passed by reference in supported PHPs, but it certainly doesn't cause any harm.

Issue #8, opened 2022-07-20
Pattern: Download button


ryelle:
Do we need a custom block for the download button, or can it be a regular button block?

Is it a static link (always `https://wordpress.org/latest.zip`) or do we need to add the version/timestamp/SHA/etc to it? In `wporg-main`, the URL is dynamic, but that's because it can be updated to point to the locale-specific build.

Current mockup:

<img width="597" alt="Screen Shot 2022-07-20 at 4 38 44 PM" src="https://user-images.githubusercontent.com/541093/180077234-d31e1903-4083-4e8d-abc7-6c749bec53fd.png">

dd32:
It seems like the immediate needs here can be solved by a shortcode, rather than a custom block. See #10 + #27.

A custom pre-defined button block with the shortcodes (and later tokens) wouldn't be out of the question though to simplify the editing experience.

ryelle:
I think we could create a pattern that uses the shortcodes, rather than a custom block, so editors can just drop the pattern in place.

ryelle:
Closing in favor of the eventual download modal block #52

PR #27, opened 2022-07-28
Add shortcodes for common dynamic numbers

### LINKED ISSUE #10: Create a shortcode for `WP_MARKET_SHARE`

dd32:
This adds ~three~ _a few_ shortcodes for various data-points:
 - `[market_share]` the value of `WP_MARKET_SHARE` including the %, defaulting to `43%`
 - `[plugin_count]` the value of `WP_PLUGIN_COUNT ` formatted to the locale number format, defaulting to `55,000`
 - `[latest_version]` the value of `WP_CORE_LATEST_RELEASE`, defaulting to the current `$wp_version` for localhost test installs. (+rosetta support)
 - `[download_link]` and `[download_link type="tar.gz"]` for the latest download link, with Rosetta support.
 - `[recommended_php]` as defined by servehappy https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/servehappy-config.php

In #10 `$meetups` is referenced, but wasn't previously displayed and doesn't appear in the [latest mockups](https://make.wordpress.org/design/2022/07/27/project-update-wordpress-org-homepage-and-download-page-mockups/), so I've skipped that. 

See #8.
Fixes #10.

Props <username>, <username>

1. Apply
2. Edit content
3. View content


dd32:
With 87853bb this should also handle the immediate need for #8 as well.

dd32:
> [recommended_php] as defined by servehappy

We might also want/need to define `[recommended_mysql_version]` and `[recommended_mariadb_version]`..

ryelle (approved):
Looks good 👍🏻

ryelle:
> We might also want/need to define [recommended_mysql_version] and [recommended_mariadb_version]

The new page content does use those values, so yeah, it would be a good addition if they're available

dd32:
>  it would be a good addition if they're available

I mean, they're not, but a shortcode that returns a static string is easy to do, and will remove the need for translation updates in the future :)

PR #33, opened 2022-07-29
Improve setup

tellyworth:
This improves the `setup:wp` step so it also runs a refresh of the content.

See #18.

1. `yarn wp-env clean all` (note this will nuke your local content!)
2. `yarn setup:wp`
3. `http://localhost:8888/` should show content matching `https://wordpress.org/main-test/`

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
 	},
 	"scripts": {
 		"setup:tools": "yarn && composer install && TEXTDOMAIN=wporg composer exec update-configs",
-		"setup:wp": "wp-env run cli \"bash env/setup.sh\"",
+		"setup:wp": "wp-env run cli \"bash env/setup.sh\" && ./env/refresh.sh",
 		"setup:refresh": "./env/refresh.sh",
 		"update:tools": "composer update && TEXTDOMAIN=wporg composer exec update-configs",
 		"wp-env": "wp-env",
--- a/readme.md
+++ b/readme.md
@@ -53,6 +53,19 @@ These must be run in the project's root folder, _not_ in theme/plugin subfolders
     yarn wp-env start
     ```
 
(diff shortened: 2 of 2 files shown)


StevenDufresne:
The code looks right, although when I get to the end of the setup, the header is not working as expected. I did run into an issue setting up:

```
  [RuntimeException]                                                                                                                              
  Failed to execute git checkout '20d67e51e9b635b142ee1ecc9585ef04f84ef4d5' -- && git reset --hard '20d67e51e9b635b142ee1ecc9585ef04f84ef4d5' --  
                                                                                                                                                  
  fatal: reference is not a tree: 20d67e51e9b635b142ee1ecc9585ef04f84ef4d5                                                                                                          
```

After deleting the composer lock and trying again, I do get a transparent header:
Screen Shot 2022-08-01 at 9 15 54 AM

package.json
```diff
@@ -13,7 +13,7 @@
 	},
 	"scripts": {
 		"setup:tools": "yarn && composer install && TEXTDOMAIN=wporg composer exec update-configs",
-		"setup:wp": "wp-env run cli \"bash env/setup.sh\"",
+		"setup:wp": "wp-env run cli \"bash env/setup.sh\" && ./env/refresh.sh",
```
StevenDufresne, line 16: You can just call the step as well:

```suggestion
		"setup:wp": "wp-env run cli \"bash env/setup.sh\" && yarn setup:refresh",
```

StevenDufresne (approved):

StevenDufresne:
> The code looks right, although when I get to the end of the setup, the header is not working as expected. I did run into an issue setting up:

I think it's related to https://github.com/WordPress/wporg-parent-2021/issues/17. You can disregard.

Issue #15, opened 2022-07-21
How do we handle translation?


tellyworth:
In the context of the front page and download pages:

* How does this relate to Rosetta?
* Which parts of those pages (if any) need to be translated via GlotPress?
* Is there content that should be translated outside GlotPress (eg by separate pages on Rosetta sites)?
* What translation mechanisms are we using in code? (see https://github.com/WordPress/wporg-main-2022/issues/1#issuecomment-1190589110)
* What other decisions do we need to make about translation?

Given the condensed time frame, we probably need a simple option that's easy, and a good option that's a post-launch v2.

dd32:
> How does this relate to Rosetta?

In an ideal world, we'd launch this for both en_US (wordpress.org/) and for all locales (xx.wordpress.org/) at the same time. However, we've got no existing successful deployments where we're translating a Block-Editor-built page, other than minimal proof-of-concepts.

> Which parts of those pages (if any) need to be translated via GlotPress?

IMHO, all visible text on the page should be translated via GlotPress. There are alternatives which might make sense in the future (Gutenberg multilingual support for example), but currently GlotPress is the defacto option here that I don't see the need to challenge.

> Is there content that should be translated outside GlotPress (eg by separate pages on Rosetta sites)?

I don't think so - however, there are some data points that might differ between English US and localised sites - For example, a locale that isn't active and their last release was an older version.  That's worth another discussion around how to handle those situations though, as we don't really want that scenario to exist.

> What translation mechanisms are we using in code?

I don't think we should do anything "special" here - just using regular `__()`/`_e()` where appropriate within any code, and using a plugin that can dynamically translate `post_content` and the other site content entered via the WordPress editor. 

That goes back to my previous mention that we don't have anything that currently works like that in production, although we have variants of it (Theme/Plugin description translations, Translated Patterns, etc) and some work that was only in concept-state for `/gutenberg/` translation.

> Given the condensed time frame, we probably need a simple option that's easy,

Whilst in an ideal world, we'd launch rosetta at the same time as the main site, realistically given the accelerated timelines in play, I feel it's safer to work with the assumption that Rosetta sites will not immediately get the new theme.
That will also ensure no rosetta/translation-related issues delay launch or any post-release issues that crop up. 

That would likely mean we'd have a rosetta-release-day as a follow-up task and enabled for an active locale (or two) for testing and as the localisation teams complete the translation for the new strings. 
[...]

PR #34, opened 2022-07-29
Add meta descriptions

### LINKED ISSUE #7: SEO: Add meta descriptions to block theme
These are defined [in code](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-main/inc/page-meta-descriptions.php).

Maybe these can be managed in WordPress itself now?

If they need to stay in code, make sure they're ported to the theme.

tellyworth:
Still needs to be updated for the new site.

Fixes #7.

1. Edit the front page post, and set the Excerpt to a known string.
2. Refresh the front page.
3. View source and confirm that your string is in the content attribute of both `<meta name="description">` and `<meta property="og:description">`
4. `yarn lighthouse` and confirm that the SEO score is 100.


dd32:
Something that comes to mind here with #15 in mind is, does hard-coding this make sense for all of this?

Should we be using a Block or other field on Posts to define the page descriptions? post_excerpt? etc?
(Oh, this is mentioned in passing on #7)

tellyworth:
Yeah at this point I'm probably going to add an old-school meta box for setting page descriptions. I hadn't though about using post_excerpt, that's a good idea.

The new way of adding a postmeta field is via a Block, but that means we'd have to somehow enforce adding that block to every page. Which seems like an oversight in core. So I figured I'd go with an old-school postmeta meta box as the simplest solution to start with. As long as there's some straightforward mechanism to translate that. Or translate the excerpt.

dd32:
> As long as there's some straightforward mechanism to translate that. Or translate the excerpt.

With what I've been working on in #15, the excerpt would be translated, and I could add a way to translate specific post_meta keys through a filter I expect.
However, there might be some other fields that need setting that don't have an easy way to define them on a per-post basis, I'm not sure..

ryelle:
> The new way of adding a postmeta field is via a Block

I'd use [PluginDocumentSettingPanel](https://developer.wordpress.org/block-editor/reference-guides/slotfills/plugin-document-setting-panel/) to insert a panel into the document sidebar, not a block. We did that for the Speaker & Session post types on WordCamp, [here's an example](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wc-post-types/js/src/session/panel-info/index.js). If you want to just get the meta fields in place in this PR, adding a better UI can happen in another PR.

tellyworth:
In the spirit of the Simplest Thing That Could Possibly Work, I've enabled `post_excerpt` on pages and used that for the meta description.

I've also set the excerpt for the front page post on the `main-test` staging site, and updated the refresh script to import the excerpt.

The excerpt is optional, it should fall back to an auto excerpt if it's not explicitly set. We could improve this later by either:

* Making the excerpt required (and perhaps re-labelling it); or
* Adding a PluginDocumentSettingPanel and switching to postmeta, as Kelly suggests.

Unless you see a problem, I'd suggest merging this as a bare minimum, then we can improve on it later when time permits.

This now gets a 100 SEO score in Lighthouse. I set the threshold to 100 for `yarn lighthouse` as a lazy regression test.

source/wp-content/themes/wporg-main-2022/inc/page-meta-descriptions.php
```diff
@@ -0,0 +1,303 @@
...
+	if ( ! $post ) {
+		return $title;
+	}
+
+	$post = get_post( $post );
+	if ( ! $post || 'page' !== $post->post_type ) {
+		return $title;
+	}
+
+	switch ( $post->page_template ) {
```
ryelle: I don't think this will work in the block-based theme, though technically all the current pages do still have this value set. Is there a reason we can't use the page title directly? I assume it was for translation in the classic theme?

source/wp-content/themes/wporg-main-2022/inc/page-meta-descriptions.php
```diff
@@ -0,0 +1,303 @@
+<?php
+/**
+ * Custom meta descriptions.
+ *
+ * @package WordPressdotorg\MainTheme
+ */
+
+namespace WordPressdotorg\Theme\Main_2022;
+
+use WordPressdotorg\API\Serve_Happy\RECOMMENDED_PHP;
```
ryelle, line 10: Is this used somewhere?

source/wp-content/themes/wporg-main-2022/inc/page-meta-descriptions.php
```diff
@@ -0,0 +1,303 @@
...
+		}
+	]
+}
+</script>
+	<?php
+}
+add_action( 'wp_head', __NAMESPACE__ . '\sites_attributes_schema' );
+
+/**
+ * Maps page titles to translatable strings.
```
ryelle: Since we'll be translating content differently than the previous theme, is this function necessary?

ryelle (approved):
This is working well - I see all the same meta tags as on production, and it uses the excerpt as description  👍🏻 

I had a few extra comments, some things I _think_ can be removed before merge, but either way this LGTM.

source/wp-content/themes/wporg-main-2022/inc/page-meta-descriptions.php
tellyworth: You're right, it's old code I'd neglected to remove. Deleted.

PR #35, opened 2022-08-01
Allow wp-cli to regenerate the .htaccess file

tellyworth:
I'm not sure if this actually solves the problem of `rewrite flush` not doing its job, but it does seem to overcome the issue with .htaccess.

`wp rewrite flush` doesn't really flush the rules. At least not in a way that reliably includes all custom rules. This might be due to the lack of apache mod_rewrite in the default wp-cli config.

0. `yarn wp-env stop && yarn wp-env start`
1.  `yarn wp-env clean all` (this will nuke all local data)
2. `yarn wp-env run cli 'rm .htaccess'`
3. `yarn setup:wp`
4. `yarn wp-env run cli 'cat .htaccess'` (check that the new .htaccess file exists)
5. Also, confirm that custom rules were flushed (how?)

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -26,6 +26,7 @@
 	"mappings": {
 		"env": "./env",
 		"wp-content/mu-plugins": "./source/wp-content/mu-plugins",
-		"wp-content/mu-plugins/0-sandbox.php": "./env/0-sandbox.php"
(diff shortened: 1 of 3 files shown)


ryelle (approved):
Works for me, and after a destroy-setup cycle, the permalinks worked 👍🏻

PR #38, opened 2022-08-02
Templates: Remove top-level block gap spacing in page content

### LINKED ISSUE #23: Remove spacing between sections
The main blocks on [the front page](https://wordpress.org/main-test/) have a default margin-block-start (top) of 24px, but this should be zero.

| Mockup | Website |
|--------|--------|
|  |  |

This is from some generated CSS, but I think this is caused by the `blockGap` property. Note that fixing this should _not_ remove the block gap entirely, since it is used for default paragraph spacing in post content.

ryelle:
Fixes #23 — this updates the `post-content` block in the template to set its `blockGap` to `0px`. 

The `blockGap` value is used to set the margins on direct descendants, keeping a consistent rhythm between elements. It can be set in the UI for many blocks, it can also be set in the templates or in theme.json, globally or per-block. In general, we want the blockGap behavior, because it adds spacing between paragraphs on normal pages. The home & download pages are built differently, so on these pages specifically, the blockGap is removed.

### Screenshots

| Before | After |
|--------|-------|
|  |  |
|  |  |

1. View the homepage
2. There should be no whitespace between sections — the colorful backgrounds should touch
3. View the download page
4. The border between the two columns should touch the light blue background below it
5. Create a new page with some regular text content, view it (you might need to short circuit the theme-switcher code)
6. There should still be space between paragraphs & other content

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/templates/front-page.html
+++ b/source/wp-content/themes/wporg-main-2022/templates/front-page.html
@@ -1,11 +1,9 @@
 <!-- wp:wporg/global-header /-->
 
-<!-- wp:query {"tagName":"main","query":"inherit"} -->
-<main class="wp-block-query">
-	<!-- wp:post-template -->
-		<!-- wp:post-content {"layout":{"inherit":true}} /-->
-	<!-- /wp:post-template -->
+<!-- wp:group {"tagName":"main"} -->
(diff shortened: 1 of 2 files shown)


StevenDufresne (approved):
Works for me!

PR #39, opened 2022-08-03
Prevent arrow characters from becoming emoji

### LINKED ISSUE #30: Arrow in CTA links is converted to emoji
In some links (the CTA style text links), the link text includes an arrow icon. The mockup uses the `↗` character, but on the staging site WordPress is converting it to ↗️  (oops, so does github).

| Mockup | Staging site |
|---------|---------|
|  |  |

It would be best if we could use the arrow character itself so it can respond to any link color.

ryelle:
Fixes #30. Text arrows added in the editor are converted to emoji, which is unexpected (and unwanted). This PR adds a filter over the content to prevent this on the set of arrows likely to be a problem (though up-right and up-left are _probably_ the only ones we'll use). It adds the text variation selector unicode character to these arrows, which forces the twemoji script to skip them, leaving them as text.

Once core fixes an issue with the `wp-exclude-emoji` class, we could switch to using that, but it would also require updating the containers for everywhere an arrow is used, so in the end this might be a better solution anyway? I've left the note about core regardless, and we can have that discussion when it's fixed.

### Screenshots

| Before | After |
|--------|-------|
|  |  |

1. Add arrows to your post content: `← ↑ → ↓ ↔ ↕ ↖ ↗ ↘ ↙`
2. The should be text arrows in the editor
3. Save your post & view it
4. The arrows should still be text arrows on the front end

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -7,6 +7,7 @@ namespace WordPressdotorg\Theme\Main_2022;
  */
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
+add_filter( 'the_content', __NAMESPACE__ . '\prevent_arrow_emoji' );
 
 /**
  * Enqueue scripts and styles.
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -7,6 +7,7 @@
  */
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
+add_filter( 'the_content', __NAMESPACE__ . '\prevent_arrow_emoji' );
```
dd32: Perhaps there should be some level of priority set here, to ensure that it runs after all other filters?

```suggestion
add_filter( 'the_content', __NAMESPACE__ . '\prevent_arrow_emoji', 20 );
```

dd32 (approved):
This seems like a safe change to me.

If we were being super-cautious, the regex should not affect characters that are followed by the variation selector, but realistically, I think it's safe to exclude that for performance reasons.

Looks good to me! 👍

Issue #40, opened 2022-08-03
Download page: Style download modal


ryelle:
~Check that the existing download modal works~ It doesn't.

Bring back the download modal, it should open after the user clicks the download button.

![Download Pop Up Message](https://user-images.githubusercontent.com/541093/182674249-16b82a45-6bd1-4612-b426-685991659ea5.png)

Also check the keyboard behavior to make sure it's accessible to keyboard-only users & screen reader users.

PR #47, opened 2022-08-04
Add download modal

adamwoodnz:
See #40

This PR adds a modal window which is displayed when the 'Download WordPress' button is clicked on the download page. The download is also triggered as it is currently.

NOTE: this component depends on the list of links style which hasn't landed yet.

### Screenshots

Screen Shot 2022-08-05 at 3 15 47 PM

1. Ensure the download button has the id `wporg__download-button`.
2. Click the Download button and the modal should be displayed. The download should also start.
3. The links in the modal should open in the same window.
4. Close the modal by clicking the close button or overlay.
5. With your keyboard tab to the button and activate with enter. The behaviour should match clicking. The modal should receive focus so that it is easy to tab to the close button or the links. These should also be able to be activated with the keyboard. Focus should cycle within the modal while tabbing. 
6. Resize the screen to check the responsiveness.


source/wp-content/themes/wporg-main-2022/src/components/download-modal/DownloadModal.js
```diff
@@ -0,0 +1,38 @@
...
+					</li>
+					<li>
+						<a href="https://www.meetup.com/pro/wordpress/">Join a local WordPress meetup ↗</a>
+					</li>
+					<li>
+						<a href="https://central.wordcamp.org/">Attend a WordCamp ↗</a>
+					</li>
+					<li>
+						<a href="https://wordpressfoundation.org/donate/">
+							Support WordPress and open source education ↗
```
adamwoodnz: Should all of this static text use translations?

source/wp-content/themes/wporg-main-2022/src/components/download-modal/DownloadModal.js
```diff
@@ -0,0 +1,38 @@
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Modal } from '@wordpress/components';
+
+const DownloadModal = ( { onClose } ) => (
+	<Modal className="wporg__download-modal" title={ __( 'Howdy!', 'wporg' ) } onRequestClose={ onClose }>
+		<>
+			<p className="wporg__download-modal-sub-header is-style-serif">Thanks for downloading WordPress</p>
```
adamwoodnz: The modal component title only accepts a string so this separate sub header section is necessary

source/wp-content/themes/wporg-main-2022/src/index.js
```diff
@@ -1 +1,19 @@
...
+	const downloadModalContainerId = 'wporg__download-modal-container';
+
+	downloadModalContainerEl.setAttribute( 'id', downloadModalContainerId );
+	document.body.appendChild( downloadModalContainerEl );
+
+	wp.element.render( <DownloadModalContainer />, document.getElementById( downloadModalContainerId ) );
+};
+
+const init = () => {
+	if ( document.body.classList.contains( 'page-template-page-download' ) ) {
```
adamwoodnz, line 14: As discussed we could do this check in `functions.php` and not load this file at all if it's not the downloads page. I'm wondering if this file will grow though as we add more pages and functionality.

source/wp-content/themes/wporg-main-2022/src/components/download-modal/DownloadModal.js
dd32: Could this text be within the Editor? To allow editors to update it without needing to have a dev update it.

That'll also resolve translating it

source/wp-content/themes/wporg-main-2022/style.css
```diff
@@ -14,3 +14,94 @@
...
+	background: rgba(30, 30, 30, 0.8);
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	padding: 20px;
+}
+
+.components-modal__frame {
+	background: var(--wp--preset--color--white);
+	max-width: 568px;
```
StevenDufresne, line 38: Where does this specific number come from? Do we need to annotate this at all?

source/wp-content/themes/wporg-main-2022/src/index.js
StevenDufresne: If that were the case, I think abstracting this block, moving it to `wporg-mu-plugins` and adding the modal content through the block editor would be a better approach. We can worry about that when we get a second modal.

StevenDufresne:
Merging this for the Demo. We can circle back and do a more in-depth review if need be.

source/wp-content/themes/wporg-main-2022/src/components/download-modal/DownloadModal.js
StevenDufresne: I updated everything to use `__` for now since this block doesn't get added to the editor right now (and may not in the future?).

source/wp-content/themes/wporg-main-2022/style.css
adamwoodnz: From Figma. It probably should be a shared dimension for all modals.

PR #49, opened 2022-08-05
Add the variation selector to the download modal arrows.

dd32:
See #40, #47.

### Screenshots

| Before | After |
|--------|-------|
|  Screen Shot 2022-08-05 at 5 12 26 pm | Screen Shot 2022-08-05 at 5 12 04 pm |

1. Apply
2. Click download button
3. Observe.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/src/components/download-modal/DownloadModal.js
+++ b/source/wp-content/themes/wporg-main-2022/src/components/download-modal/DownloadModal.js
@@ -19,19 +19,19 @@ const DownloadModal = ( { onClose } ) => (
 				</p>
 				<ul>
 					<li>
-						<a href="https://make.wordpress.org/">{ __( 'Get involved in WordPress ↗', 'wporg' ) }</a>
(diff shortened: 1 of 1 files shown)


adamwoodnz:
There is no discernible change in the diff, what did you change exactly please?

ryelle:
The text variation selector, `&#xFE0E;` was added - it's an invisible character, it only exists to indicate that the character before it should be displayed as text, not an emoji. It prevents the twemoji script in WordPress from converting the arrow into ↗️.

If you encoded the diff, you'd see it:

```diff
- { __( 'Get involved in WordPress &#x2197;', 'wporg' ) }
+ { __( 'Get involved in WordPress &#x2197;&#xFE0E;', 'wporg' ) }
```

There's some more discussion in this issue https://github.com/WordPress/wporg-main-2022/issues/30 & PR https://github.com/WordPress/wporg-main-2022/pull/39

dd32:
Here's another way to view it, `git diff` but with `LC_ALL=C` env set which forces the display of the raw multibyte characters (As `C` is effectively ascii, a-z0-9etc only)

Screen Shot 2022-08-09 at 3 30 11 pm

PR #53, opened 2022-08-08
Try/save post content

tellyworth:
This will save the contents of posts downloaded from the staging site in local json files, in `./env/post-data`.

The idea is to potentially maintain an archive of content in case of data loss etc. This may not be necessary, but it's very easy to implement. Posting this as a draft PR for feedback.

To use, just run `yarn setup:refresh`.


tellyworth:
Not needed, the solution to #55 will probably achieve something similar.

PR #54, opened 2022-08-08
Add hreflang link tags

### LINKED ISSUE #43: Add hreflang tags
The new pages need `hreflang` tags as a critical SEO requirement: https://wordpress.slack.com/archives/C02QB8GMM/p1659514993386139?thread_ts=1658956086.933959&cid=C02QB8GMM

The old front page used a static list. Some context on that: https://meta.trac.wordpress.org/ticket/5108

For other pages(?) it used this quirky code: https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg/functions.php#L187

tellyworth:
This is basically a straight port of the `hreflang_link_attributes` function from the `wporg` theme.

I made a couple of quick edits so it runs in local testing without GlotPress installed.

Fixes #43.

Props @dd32.

1. Refresh the main page.
2. View source.
3. Confirm the presence of at least two `<link rel="alternate" href="..." hreflang="..">` elements.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/inc/hreflang.php
+++ b/source/wp-content/themes/wporg-main-2022/inc/hreflang.php
@@ -1,0 +1,163 @@
+<?php
+
+namespace WordPressdotorg\Theme\Main_2022;
+
+/**
+ * Adds hreflang link attributes to WordPress.org pages.
+ *
+ * @link https://support.google.com/webmasters/answer/189077?hl=en Use hreflang for language and regional URLs.
+ * @link https://sites.google.com/site/webmasterhelpforum/en/faq-internationalisation FAQ: Internationalisation.
+ */
+function hreflang_link_attributes() {
+	// No hreflangs on 404 pages.
+	if ( is_404() ) {
+		return;
+	}
+
+	if ( ! defined( 'GLOTPRESS_LOCALES_PATH' ) && 'local' !== wp_get_environment_type() ) {
+		return;
+	}
+
+	// Don't add these if within a rest-api request which calls `wp_head`. It's never going to be expected.
+	if ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
+		return;
+	}
+
+	wp_cache_add_global_groups( array( 'locale-associations' ) );
+
+	// Google doesn't have support for a whole lot of languages and throws errors about it,
+	// so we exclude them, as we're otherwise outputting data that isn't used at all.
+	$unsupported_languages = array(
+		'arq',
+		'art',
+		'art-xemoji',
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-main-2022/inc/hreflang.php
```diff
@@ -0,0 +1,163 @@
...
+			'locale'    => 'x-default',
+			'hreflang'  => 'x-default',
+			'subdomain' => '',
+		);
+
+		uasort( $sites, function( $a, $b ) {
+			return strcasecmp( $a->hreflang, $b->hreflang );
+		} );
+
+		wp_cache_set( 'local-sites', $sites, 'locale-associations' );
```
tellyworth, line 144: I thought about making the cache_set conditional on `count( $sites ) > 2` so my local testing bypass can't poison the cache in production if it somehow executed the wrong code path.

source/wp-content/themes/wporg-main-2022/inc/hreflang.php
```diff
@@ -0,0 +1,166 @@
...
+		'tzm',
+	);
+
+	$sites = wp_cache_get( 'local-sites', 'locale-associations' );
+
+	// WARNING: for any changes below, check other uses of the `locale-assosciations` group as there's shared cache keys in use.
+	if ( false === $sites ) {
+		global $wpdb;
+
+		if ( 'local' !== wp_get_environment_type() ) {
```
dd32: I would be tempted to write this somewhat like this, just to reduce the indentation a level.

```
if ( 'local' === wp_get_environment_type() ) {
   $sites = array();
} else {
   $sites = $wpdb...
   require ....
}

foreach ( $sites as ....
```

source/wp-content/themes/wporg-main-2022/inc/hreflang.php
```diff
@@ -0,0 +1,166 @@
...
+	if ( is_singular() ) {
+		$path = parse_url( get_permalink(), PHP_URL_PATH );
+	} else {
+		// WordPress doesn't have a good way to get the canonical version of non-singular urls.
+		$path = $_SERVER['REQUEST_URI']; // phpcs:ignore
+	}
+
+	foreach ( $sites as $site ) {
+		$url = sprintf(
+			'https://%swordpress.org%s',
```
dd32, line 156: I guess we don't really care that this will reference `wordpress.org` as the hreflang when it's being used locally, so I see no problem keeping the production domain here.

tellyworth:
I think this is good to go now but let's wait till after the DC review.

PR #56, opened 2022-08-08
Add smoke tests

tellyworth:
This adds some basic smoke tests (mis)using phpunit.

The tests will fetch `http://localhost:8888` and `http://localhost:8888/download/` and run a few basic sanity tests on the HTML output.

Sample output:

```bash
% yarn test
yarn run v1.22.18
$ ./vendor/phpunit/phpunit/phpunit tests/phpunit
PHPUnit 7.5.20 by Sebastian Bergmann and contributors.

........SS                                                        10 / 10 (100%)

Time: 6.05 seconds, Memory: 4.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 10, Assertions: 12, Skipped: 2.
✨  Done in 6.17s.
```

1. `yarn test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
 		"backstop:test": "npm exec backstop test",
 		"lighthouse": "lighthouse-ci http://localhost:8888/ --accessibility=100 --best-practices=100 --seo=100",
 		"lighthouse:desktop": "lighthouse http://localhost:8888/ --view --preset=desktop --output-path=lighthouse.html",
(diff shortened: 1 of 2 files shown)


ryelle:
Is this meant to be a framework for other tests? It works for me, but I'm not totally sure why we need it.

tellyworth:
Intended as a starting point, and potentially a framework for more tests if it's useful.

It probably isn't useful though if there's not a good way to incorporate it into the sync/deploy process. Probably belongs in next iteration / pending discussion, to figure out if it makes sense.

Issue #52, opened 2022-08-08
Convert download modal to block


adamwoodnz:
![Screen Shot 2022-10-11 at 10 32 20 AM](https://user-images.githubusercontent.com/1017872/194955496-54740a64-f90f-4dcc-a8a0-dbc2f10d6006.jpg)

The modal on the Download page is currently custom coded:
Initialisation: https://github.com/WordPress/wporg-main-2022/blob/trunk/source/wp-content/themes/wporg-main-2022/src/download/index.js
Component: https://github.com/WordPress/wporg-main-2022/blob/trunk/source/wp-content/themes/wporg-main-2022/src/components/download-modal/download-modal.js

### Problems

1. The content is hard coded, and the component isn't reusable in other pages without further development.
2. Loading the modal component from Gutenberg on the frontend with `import { Modal } from '@wordpress/components';` loads the entire `@wordpress/components` package on the client. If we could render the modal server side using a block we could avoid this.

### Requirements

All content configured through the page editor. Input fields required to achieve this:

1. Title (text)
3. Subtitle (text)
4. Content (rich text)

To make the component more reusable we could consider adding button config/content for actions, eg. 'OK', 'Dismiss', or some other call to action.

The modal block would need to be linked to a button or link block in the page content somehow.

The block should probably live in https://github.com/WordPress/wporg-mu-plugins

ryelle:
I've moved this out of the current iteration since we don't need it for launch, but I do think it's a good idea for the future. There will probably be other places we'll need a content modal (for example, we have some on the Patterns site).

Tagging it Needs Design so we can work with designers to figure out the editor interface.

Probably needs a button which can toggle the modal, and then an area for the writer to enter modal content. The button should have all the same features as the core button block. What else does it need?

jasmussen:
> Tagging it Needs Design so we can work with designers to figure out the editor interface.

Sorry I missed this. Can you expand a little bit on the context? Which page will the modal be invoked from, what's the contents of the modal, and what interface will be necessary for the editor block? Thank you 🙏

adamwoodnz:
Hi @jasmussen I've added more detail in the description. Let me know if anything doesn't make sense 🙏

jasmussen:
Ah, thank you that's good context. Not related to this issue, but just seeing the external link indicators next to the links in the modal, are those icons necessary in a webpage where presumably everything is an "external link"?

Regarding the modal:

[...]

PR #59, opened 2022-08-08
Only load the download modal script if on the download page

### LINKED ISSUE #58: Performance: only load modal scripts on Download page
The Download page modal script uses Gutenberg Modal which loads the entire `wordpress/components` package. It is designed not to run on other pages however this dependency is loaded regardless.

For now we can't load only the code for the Modal so as a workaround we want to only load this script on the Download page. 

**To Reproduce**
1. Go to your local home page
2. Using your devtools observe the `wporg-main-2022-script` script tag at the bottom of the body, and in the network tab or code coverage observe http://localhost:8888/wp-content/plugins/gutenberg/build/components/index.min.js being loaded

adamwoodnz:
Fixes #58

The download modal script loads `wordpress/components` which is large and blocks page rendering. While we can't load the code only for the modal for now, we can choose to only load this dependency on the download page, thereby improving performance of the home page.

1. Load your local home page
2. Check that the `wporg-main-2022-script` is not loaded and there is no request for http://localhost:8888/wp-content/plugins/gutenberg/build/components/index.min.js
3. Navigate to the download page
4. Click the Download WordPress button and the modal should still appear

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -25,22 +25,24 @@ function enqueue_assets() {
 		filemtime( __DIR__ . '/style.css' )
 	);
 
-	$path        = __DIR__ . '/build/index.js';
-	$deps_path   = __DIR__ . '/build/index.asset.php';
-	$script_info = file_exists( $deps_path )
-		? require $deps_path
-		: array(
-			'dependencies' => array(),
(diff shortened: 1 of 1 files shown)


dd32 (approved):
For now this looks fine to me.

Noting that
 - The theme JS is currently only this DownloadModal code, and
 - #52 will ultimately remove this

PR #60, opened 2022-08-08
Theme: Override spacing values on small screens

### LINKED ISSUE #12: Small screen layout & typography
This is not a specific issue, but something to pay attention to. GB doesn't let you set different spacing or font size based on screen sizes (though I remember an issue about this, so we should confirm), so getting this to look nice on smaller screens will probably require CSS.

I've added the "Needs design" label since we can't do anything here until there is a finished mockup 🙂

ryelle:
Fixes #12 — This adds a bunch of overrides to the theme's CSS to adjust the spacing on smaller screens. All the edits should be documented inline, but in general I'm adding left & right padding so the content isn't touching the edges and adjusting the column gap spacing when the columns switch to a single-col layout.

### Screenshots

Home

| Size | Before | After |
|-----|--------|-------|
| 1280px | trunk-home-1280 | ss-home-1280 |
| 780px | trunk-home-780  | ss-home-780 |
| 480px | trunk-home-480 | ss-home-480 |

Download

| Size | Before | After |
|-----|--------|-------|
| 1280px | trunk-download-1280 | ss-download-1280 |
| 780px | trunk-download-780 | ss-download-780 |
| 480px | trunk-download-480 | ss-download-480 |

1. View on different screen sizes.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/style.css
+++ b/source/wp-content/themes/wporg-main-2022/style.css
@@ -15,13 +15,135 @@
  * templates or theme.json settings.
  */
 
-/* 
+/*
+ * Override inline paddings with the spacing preset.
+ */
+
+[style*="padding-top:120px"] {
+	padding-top: var(--wp--preset--spacing--80) !important;
+}
+
+[style*="padding-bottom:120px"] {
+	padding-bottom: var(--wp--preset--spacing--80) !important;
+}
+
+[style*="padding-top:160px"] {
+	padding-top: var(--wp--preset--spacing--90) !important;
+}
+
+[style*="padding-bottom:160px"] {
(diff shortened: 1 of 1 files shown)


ryelle:
I'm going to merge this now so that it can be deployed to /main-test/, if there are any issues we can fix those in followup PRs.

adamwoodnz:
Looks like we may have missed the paragraph below Get Started on home?

Screen Shot 2022-08-10 at 8 36 12 AM

ryelle:
I ignored that section because there's a fix for the arrow over here: https://github.com/WordPress/wporg-parent-2021/pull/36 -- I missed that the text is also misaligned, but I'll fix it after the arrow fix merges.

PR #64, opened 2022-08-11
Theme: Add page content as patterns

ryelle:
This PR pulls the content out of the editor and puts it into patterns, so that it can be under version control. Each pattern can then be loaded using the `pattern` block, an internal block that renders a block pattern. The patterns are just the same content that was in the post editor.

I decided to go with a pattern, rather than dropping the content into the HTML template, so that we can use PHP as needed (ex, for translation functions). I haven't added those yet, though.

In the future, we could try some kind of syncing system to copy down page content from a WP install?

Another iteration could look at hiding these from the site/page editor, so they can only be added via code.

1. View the branch
2. Make a change in the page content in the editor, like, delete everything
3. The page will still look the same
4. Make a change in the pattern code
5. It will be reflected on the front end

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/front-page.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/front-page.php
@@ -1,0 +1,283 @@
+<?php
+/**
+ * Title: Front Page
+ * Slug: wporg-main-2022/front-page
+ * Categories:
+ */
+
+?>
+<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"120px","right":"0px","bottom":"120px","left":"0px"},"blockGap":"30px"}},"layout":{"inherit":false}} -->
+<div class="wp-block-group alignwide" id="intro" style="padding-top:120px;padding-right:0px;padding-bottom:120px;padding-left:0px"><!-- wp:columns {"style":{"spacing":{"blockGap":"0px"}}} -->
+<div class="wp-block-columns"><!-- wp:column {"width":"85%"} -->
+<div class="wp-block-column" style="flex-basis:85%"><!-- wp:heading {"level":1,"textColor":"charcoal-1","fontSize":"heading-cta"} -->
+<h1 class="has-charcoal-1-color has-text-color has-heading-cta-font-size">WordPress: Publish your <em><mark class="has-inline-color has-blueberry-1-color">passion</mark></em></h1>
+<!-- /wp:heading --></div>
+<!-- /wp:column -->
+
+<!-- wp:column {"width":"15%"} -->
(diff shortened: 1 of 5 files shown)


StevenDufresne:
Test case passes on my machine!

source/wp-content/themes/wporg-main-2022/patterns/front-page.php
```diff
@@ -0,0 +1,283 @@
...
+<div class="wp-block-button"><a class="wp-block-button__link wp-element-button" href="/download">Get WordPress</a></div>
+<!-- /wp:button --></div>
+<!-- /wp:buttons --></div>
+<!-- /wp:group -->
+
+<!-- wp:wporg/language-suggest {"align":"full"} -->
+<div class="wp-block-wporg-language-suggest alignfull"></div>
+<!-- /wp:wporg/language-suggest -->
+
+<!-- wp:cover {"url":"https://wordpress.org/main-test/files/2022/08/editor-bg-scaled.webp","id":9215,"dimRatio":0,"minHeight":670,"minHeightUnit":"px","isDark":false,"align":"full","style":{"spacing":{"padding":{"top":"80px","right":"0px","bottom":"80px","left":"0px"}}}} -->
```
tellyworth, line 37: We'll need to update image URLs when we move to the live site, but if anything this makes it easier.

source/wp-content/themes/wporg-main-2022/patterns/front-page.php
```diff
@@ -0,0 +1,283 @@
...
+<!-- /wp:column --></div>
+<!-- /wp:columns --></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"margin":{"top":"30px","bottom":"30px"}}},"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignwide" style="margin-top:30px;margin-bottom:30px"><!-- wp:image {"id":9099,"width":105,"height":33,"sizeSlug":"full","linkDestination":"none"} -->
+<figure class="wp-block-image size-full is-resized"><img src="https://wordpress.org/main-test/files/2022/08/Spotify.png" alt="" class="wp-image-9099" width="105" height="33"/></figure>
+<!-- /wp:image -->
+
+<!-- wp:image {"id":9100,"width":79,"height":33,"sizeSlug":"full","linkDestination":"none"} -->
```
tellyworth, line 137: Likewise for updating attachment IDs. It doesn't break anything to launch as-is, so that's not a blocker, just something to follow up on later.

tellyworth (approved):
Brilliant solution! Super simple and it works well.

PR #72, opened 2022-08-12
Stop lazy loading of content images above the fold to improve LCP

### LINKED ISSUE #45: Performance: front-end
Let's put together a checklist of any front-end performance issues (current or potential) that need to be considered. We may not fix them all for initial launch, but the sooner we know about them the better.

- [x] Use webp for images
- [x] Load the download modal JS only on download page
- [x] Load the MU Latest Posts block JS in the footer
- [x] Largest Contentful Paint image was lazy loaded

adamwoodnz:
Partially fixes #45 

Content images which are above the fold shouldn't be lazy loaded as it affects the Largest Contentful Paint of the page, a key performance metric.

Screen Shot 2022-08-12 at 2 50 17 PM

This PR adds a filter to skip X number of images as outlined on https://make.wordpress.org/core/2021/12/29/enhanced-lazy-loading-performance-in-5-9/, however the filter doesn't fire, @tellyworth suggested perhaps related to the content being in a pattern. Not sure if this being page content rather than post content makes a difference either.

The image in question is actually the first content image in the page, so if I understand correctly it should not be lazy loaded by default. It seems we're doing something differently to the way this optimization expects.

### Screenshots

| Before | After |
|--------|-------|
| image  | image |

1.
2.
3.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -11,6 +11,7 @@ require_once( __DIR__ . '/inc/hreflang.php' );
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
 add_filter( 'render_block_core/pattern', __NAMESPACE__ . '\prevent_arrow_emoji', 20 );
+add_filter( 'wp_omit_loading_attr_threshold', __NAMESPACE__ . '\skip_lazyloading_on_homepage_images' );
 
 /**
  * Enqueue scripts and styles.
@@ -107,3 +108,14 @@ add_action(
 		);
 	}
 );
+
+/**
+ * Prevent lazy loading the images above the fold as this affects LCP performance.
+ */
+function skip_lazyloading_on_homepage_images( $omit_threshold ) {
+	if ( is_home() ) {
+		return 2;
+	}
+
+	return $omit_threshold;
+}
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -11,7 +11,7 @@ require_once( __DIR__ . '/inc/hreflang.php' );
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
 add_filter( 'render_block_core/pattern', __NAMESPACE__ . '\prevent_arrow_emoji', 20 );
-add_filter( 'wp_omit_loading_attr_threshold', __NAMESPACE__ . '\skip_lazyloading_on_homepage_images' );
+add_filter( 'wp_img_tag_add_loading_attr', __NAMESPACE__ . '\override_lazy_loading', 10, 2 );
 
 /**
  * Enqueue scripts and styles.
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -93,3 +94,14 @@ function() {
 		return new \WP_Error();
 	}
 );
+
+/**
+ * Prevent lazy loading the images above the fold as this affects LCP performance.
+ */
+function skip_lazyloading_on_homepage_images( $omit_threshold ) {
+	if ( is_home() ) {
+		return 2;
```
adamwoodnz: Draft value

ryelle:
The first issue I ran into is that the lazy attribute is only added to images on the current site, so my local site wasn't adding it because it was pulled from `main-test`. I had to add it to my local site & swap the pattern code. In case anyone else runs into that :)

I tracked how that threshold value is used, and it's only used if the image is in the content, otherwise it defaults to `lazy`. In our case, context is `false`, so we return before even trying the threshold.

```php
// Only elements with 'the_content' or 'the_post_thumbnail' context have special handling.
if ( 'the_content' !== $context && 'the_post_thumbnail' !== $context ) {
	return 'lazy';
}
```
[(source)](https://github.com/WordPress/wordpress-develop/blob/c45ea397da93cbe3536e4fb1c8f94abc6300b305/src/wp-includes/media.php#L5496-L5499)

So instead, I tried filtering the attribute itself when it's added, [with `wp_img_tag_add_loading_attr`.](https://github.com/WordPress/wordpress-develop/blob/c45ea397da93cbe3536e4fb1c8f94abc6300b305/src/wp-includes/media.php#L2036-L2049) This doesn't give us any context, so instead I'm checking for the file name and disabling lazy-loading for this specific image. We could check for other things, like the class name, but then that would affect all cover blocks…

The value of `$image` is just the image tag (whitespace added for readability).
```
<img
	width="2560"
	height="1305"
	class="wp-block-cover__image-background wp-image-9084"
	alt=""
	src="http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1.webp"
	data-object-fit="cover"
	srcset="
		http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1.webp           2560w,
		http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1-300x153.webp    300w,
		http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1-1024x522.webp  1024w,
		http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1-768x392.webp    768w,
		http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1-1536x783.webp  1536w,
		http://localhost:8888/wp-content/uploads/2022/08/editor-bg-scaled-1-2048x1044.webp 2048w
	"
	sizes="(max-width: 2560px) 100vw, 2560px"
/>
```

Anyway, I think https://github.com/WordPress/wporg-main-2022/commit/e7a6e6509cbafe9fac1c293b84319d21d17b4ced is a workaround we can launch with, and we can think about a future-safe solution later.

ryelle:
Lighthouse says the Largest Contentful Paint is now another section, and it doesn't flag it.

Screen Shot 2022-08-12 at 1 50 47 PM

tellyworth:
> So instead, I tried filtering the attribute itself when it's added, [with `wp_img_tag_add_loading_attr`.](https://github.com/WordPress/wordpress-develop/blob/c45ea397da93cbe3536e4fb1c8f94abc6300b305/src/wp-includes/media.php#L2036-L2049) This doesn't give us any context, so instead I'm checking for the file name and disabling lazy-loading for this specific image.

Nice one Kelly! That's exactly where I was trying to get to, I hadn't found that particular filter though.

This is a good workaround for now, @adamwoodnz let's merge it if it works.

As a more flexible solution going forward, instead of string-matching the file name, could we use something like a special classname or other attribute on the image to trigger it? That way it could be more explicitly controlled in the template rather than hidden away in a magic filter. Definitely not a blocker for merging though.

adamwoodnz:
> This is a good workaround for now, @adamwoodnz let's merge it if it works.

👍 testing in progress

adamwoodnz (commented):
LGTM, couldn't get a local image working, but tested in Sandbox

adamwoodnz:
> could we use something like a special classname or other attribute on the image to trigger it

I think we could apply a `image-no-lazy` or similar class to the image in the pattern, but it's not possible to add a class directly to the image through the editor as far as I can see. Couldn't get this to work in my sandbox as the pattern files don't see to get used like they do locally.

PR #73, opened 2022-08-12
Missing entry-content class

tellyworth:
Co-Authored-By: Steven Dufresne <dufresnesteven@gmail.com>

Fixes a bug in mobile padding introduced in #64 

### Screenshots

| Before | After |
|--------|-------|
| Screen Shot 2022-08-12 at 3 29 47 pm |Screen Shot 2022-08-12 at 3 31 00 pm |

Ignore the missing lang banner.

1.
2.
3.


StevenDufresne (approved):
I'm not sure if this is the most future-proof way of doing it, but it does in fact restore padding!

carolinan:
Will this not cause a block validation error? The class name is not in the comment, only on the element?

ryelle:
Looks like it doesn't cause an error, actually, but we should add it in the block comment too.

PR #75, opened 2022-08-12
Add structured content to download page

### LINKED ISSUE #5: Downloads: Add structured data in block theme
This code is currently added in the `page-download.php` template directly, so it'll need to be moved somewhere else in the theme.

[Source of current code.](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-main/page-download.php#L69)

ryelle:
Fixes #5 — This adds in the structured content from the old theme to the download page.

I also added a shortcode for the version time, to match how the version and URL are managed. It can be used with a `format` attribute, like so: `[latest_version_date format="Y-m-d" /]`. It defaults to the `Y-m-d\TH:i:s\+00:00` format.

1. Open the download page, view source
2. You should see a section with `<script type="application/ld+json">` and some JSON, make sure it matches the code currently on `wordpress.org/download`
3. Open the homepage, view source
4. Confirm the `<script type="application/ld+json">` code is still on the page (no change from trunk)

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
+++ b/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
@@ -95,3 +95,24 @@ add_shortcode(
 		return $link;
 	}
 );
+
+/**
+ * Shortcode for a formatted time when the latest version was published.
+ */
+add_shortcode(
+	'latest_version_ts',
+	function( $attrs = array() ) {
+		$format = $attrs['format'] ?? 'Y-m-d\TH:i:s\+00:00';
+
+		$timestamp = defined( 'WPORG_WP_RELEASES_PATH' ) ? filemtime( WPORG_WP_RELEASES_PATH . 'wordpress-' . WP_CORE_LATEST_RELEASE . '.zip' ) : time();
+
+		if ( defined( 'IS_ROSETTA_NETWORK' ) && IS_ROSETTA_NETWORK && ! empty( $GLOBALS['rosetta'] ) ) {
+			$rosetta_release = $GLOBALS['rosetta']->rosetta->get_latest_public_release();
+			if ( $rosetta_release ) {
+				$timestamp = $rosetta_release['builton'];
+			}
+		}
+
+		return gmdate( $format, $timestamp );
+	}
+);
--- a/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
+++ b/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
@@ -97,10 +97,10 @@ add_shortcode(
 );
 
 /**
- * Shortcode for a formatted time when the latest version was published.
+ * Shortcode for a formatted date & time when the latest version was published.
  */
 add_shortcode(
-	'latest_version_ts',
+	'latest_version_date',
 	function( $attrs = array() ) {
 		$format = $attrs['format'] ?? 'Y-m-d\TH:i:s\+00:00';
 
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
```diff
@@ -95,3 +95,24 @@ function( $attrs = array() ) {
...
+
+/**
+ * Shortcode for a formatted date & time when the latest version was published.
+ */
+add_shortcode(
+	'latest_version_date',
+	function( $attrs = array() ) {
+		$format = $attrs['format'] ?? 'Y-m-d\TH:i:s\+00:00';
+
+		$timestamp = defined( 'WPORG_WP_RELEASES_PATH' ) ? filemtime( WPORG_WP_RELEASES_PATH . 'wordpress-' . WP_CORE_LATEST_RELEASE . '.zip' ) : time();
```
iandunn, line 107: Is the fallback here for local envs? If so, it might be safer to do something like:

```
if env == production {
    use timestamp of file
} else {
   use time()
}
```

That way it won't fail silently in production and output the wrong time if something goes wrong in the future.

That's subjective, though, and definitely not a blocker.

source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
```diff
@@ -95,3 +95,24 @@ function( $attrs = array() ) {
...
+
+		if ( defined( 'IS_ROSETTA_NETWORK' ) && IS_ROSETTA_NETWORK && ! empty( $GLOBALS['rosetta'] ) ) {
```
iandunn, line 109: `is_rosetta_site()` (from the global header) may provide the same result, but i haven't checked if the differences in implementation would have any side effects

iandunn (commented):
The homepage and download page have the expected output in my tests. Other pages don't have any schema 👍🏻 

Left some minor comments, but feel free to close them and merge if you disagree.

source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
dd32: `IS_ROSETTA_NETWORK` was copied from the other shortcodes.

https://github.com/WordPress/wporg-main-2022/blob/trunk/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php#L67

I used the constant in these cases as checking the specific class-name of the variable seemed less forward-compatible when/if the plugins change. 

`is_rosetta_network()` or this really has no major differences that I'm aware of, if we want to switch to a singular global function, I'd be okay with that!

source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
dd32: Personally, I would be more comfortable checking the constant than the environment type, although in the event that the file doesn't exist this would be `false` anyway.

Although, as the fallback here, I wouldn't be against just returning nothing instead of the current time, as if there's no the appropriate data, it doesn't seem proper to show it.. not that it really matters much IMHO

```suggestion
		$timestamp = false;
		if ( defined( 'WPORG_WP_RELEASES_PATH' ) ) {
			$timestamp = filemtime( WPORG_WP_RELEASES_PATH . 'wordpress-' . WP_CORE_LATEST_RELEASE . '.zip' );
		}
```
plus, then replace `return gmdate( $format, $timestamp );` with something like `return $timestamp ? gmdate( $format, $timestamp ) : '';`

dd32 (approved):

dd32:
Just noting, this can be merged and the above review comments resolved afterwards IMHO.

PR #82, opened 2022-08-15
Replace image URLs with CDN

tellyworth:
This uses CDN image URLs for front page images.

1. Test before/after - should be no visual difference
2. Make sure images on the front page have `https://s.w.org/` URLs instead of `https://wordpress.org/`
3.


tellyworth:
I think this is probably unnecessary given that we're using Photon on the main site. 

Disregard unless needed.

ryelle:
Looks like Photon is enabled on production, and the images are CDN'd. If we're OK with the wpcom CDN, `https://i1.wp.com/…`, then I think we can close this.

PR #93, opened 2022-08-17
Import attachment posts from staging

tellyworth:
This imports attachment posts to the local env. Possible solution to the different image behaviour we've been seeing on local vs production.

WIP.

### DIFF
--- a/env/import-content.php
+++ b/env/import-content.php
@@ -58,7 +58,7 @@ function filter_curl_options( $ch ) {
  * Download a remote URL and sideload it into the upload directory.
  * Doesn't create an attachment.
  */
-function sideload_from_url( $remote_url ) {
+function sideload_from_url( $remote_url, $time = null ) {
 	require_once( ABSPATH . 'wp-admin/includes/file.php' );
 
 	$temp_file = download_url( $remote_url );
@@ -78,7 +78,7 @@ function sideload_from_url( $remote_url ) {
 		'test_form'   => false,
 	);
 
-	return wp_handle_sideload( $file_array, $overrides );
+	return wp_handle_sideload( $file_array, $overrides, $time );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tellyworth:
@dd32 got a sec for an opinion here?

The main reason for doing this is so that we have attachment posts with matching IDs, so there's (hopefully) identical behaviour from blocks running locally as from prod.

With respect to the files themselves, I'm undecided whether it should:

* Also download the media files and store them locally, with local attachment post metadata pointing to the local files (easy but doesn't quite match prod); or
* Don't download the files, and have the local attachment post metadata point to the remote file URLs (not sure if this easy or reliable)

What's your take?

dd32:
> Also download the media files and store them locally, with local attachment post metadata pointing to the local files (easy but doesn't quite match prod); or

Assuming that the IDs match between local and prod, this doesn't seem like a terrible idea. It would also allow for image-editing functions (and others) to be used on the local copy.. It would not necessarily require updating the post contents to use the local image URLs though, from my understanding it should just work if it's got the remote urls in the post content still?

> Don't download the files, and have the local attachment post metadata point to the remote file URLs (not sure if this easy or reliable)

But this also sounds like a potential solution, but it would probably require a plugin to filter the URLs, that would then require ongoing maintenance, as I believe WordPress has moved away from storing absolute paths (on disk, and http urls) as part of the media metadata. That would make this a potentially more complicated and more likely to have side-effects. One thing that comes to mind here, is that there's some work being done around webp's in core that would expect the source-files to exist on disk for on-the-fly/background-processing conversion.

tellyworth:
I went with the first option, thanks for the advice. 

@ryelle this seems to work ok but isn't thoroughly tested.  I'm not entirely clear on whether or not it solves the issue of (for ex) different lazy loading behaviour on local.

PR #100, opened 2022-08-22
Build pattern files from post content

### LINKED ISSUE #55: Editing safety in production
Once the site is launched, anyone editing the front page will face a situation where they can't preview their changes or save a draft, because the page is already published. Effectively they're editing "live".

How can we make that a safer environment?

Some possibilities:
* Lock the publish button and require some special process for changing the content. 
* Use the https://github.com/WordPress/revisions-extended plugin
* Use some other workflow plugin
* ..?

One actionable step would be to test out revisions-extended and see if it works and is suitable.

Other ideas?

tellyworth:
This adds a new `build:patterns` script that will generate updated pattern files in the theme, based on edited post content.

It includes some hard-coded parts initially, this is a starting point with some decisions still to make. For example:

* Currently it treats `w.org/main-test/` as the canonical source of content. Should that become w.org itself, or a staging site?
* Should it directly fetch the remote content (as it does now) or rely on 'setup:refresh' and use the local version of content?
* Which of those will make translation easier?
* Where should the URLs and corresponding output files be defined?
* How could it programatically discover newly created pages on the staging/canonical site?
* Should this be done locally by devs, or as part of the sandbox sync?

To use/test:

```bash
yarn build:patterns
```

..then `git diff` and check that it has generated updated files in `source/wp-content/themes/wporg-main-2022/patterns`.

### DIFF
--- a/env/export-content.php
+++ b/env/export-content.php
@@ -55,6 +55,6 @@ EOF;
 	if ( false === $bytes ) {
 		die( 'Unable to write to ' . $output_path );
 	} else {
-		echo 'Wrote ' . number_format( $bytes ) . ' to ' . $output_path . "\n";
+		echo 'Wrote ' . number_format( $bytes ) . ' bytes to ' . $output_path . "\n";
 	}
 }
--- a/env/export-content.php
+++ b/env/export-content.php
@@ -1,0 +1,60 @@
+#!/usr/bin/php
+<?php
+// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+
+namespace WordPress_org\Main_2022\ImportTestContent;
+
+/**
+ * CLI script for exporting post content from the local DB to local pattern/template files.
+ *
+ * Intended to be run after import-content.php fetches the latest content from a remote prod/staging server.
+ *
+ * This is a vanilla PHP script, not a WP script.
+ *
+ */
+
+// This script should only be called in a CLI environment.
+if ( 'cli' != php_sapi_name() ) {
+	die();
+}
+
+$opts = getopt( 'u:o:', array( 'url:', 'output:' ) );
+
+$url = $opts['u'] ?? $opts['url'] ?? null;
+$output = $opts['o'] ?? $opts['output'] ?? null;
+
+if ( $url && $output ) {
+
+	$data = file_get_contents( $url );
+	if ( !$data ) {
+		die( "Unable to fetch {$url}\n" );
+	}
+
+	$posts = json_decode( $data );
+	$post = $posts[0] ?? null;
+	if ( !isset( $post->content_raw ) ) {
+		var_dump( $post );
+		die( "No content_raw available at {$url}\n" );
+	}
+
+	$header = <<<EOF
+<?php
+/**
+ * Title: {$post->title->rendered}
+ * Slug: wporg-main-2022/{$post->slug}
+ * Categories:
+ */
+
+?>
+
+EOF;
+
+	$output_path = $output;
+	$bytes = file_put_contents( $output_path, $header . $post->content_raw . "\n" );
+
+	if ( false === $bytes ) {
+		die( 'Unable to write to ' . $output_path );
+	} else {
+		echo 'Wrote ' . number_format( $bytes ) . ' to ' . $output_path . "\n";
+	}
+}
--- a/env/export-content.php
+++ b/env/export-content.php
@@ -18,12 +18,15 @@ if ( 'cli' != php_sapi_name() ) {
 	die();
 }
 
-$opts = getopt( 'u:o:', array( 'url:', 'output:' ) );
+$opts = getopt( 'u:o:', array( 'url:', 'output:', 'manifest:', 'rest_url:', 'pattern_path:' ) );
 
 $url = $opts['u'] ?? $opts['url'] ?? null;
 $output = $opts['o'] ?? $opts['output'] ?? null;
+$manifest = $opts['manifest'] ?? null;
+$rest_url = $opts['rest_url'] ?? 'http://wordpress.org/main-test/wp-json/wp/v2/pages?context=wporg_export&slug=%s';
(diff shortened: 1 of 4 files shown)


env/export-content.php
```diff
@@ -0,0 +1,60 @@
...
+
+EOF;
+
+	$output_path = $output;
+	$bytes = file_put_contents( $output_path, $header . $post->content_raw . "\n" );
+
+	if ( false === $bytes ) {
+		die( 'Unable to write to ' . $output_path );
+	} else {
+		echo 'Wrote ' . number_format( $bytes ) . ' to ' . $output_path . "\n";
```
ryelle: This confused me for a second because I thought it was post IDs. Maybe add
```suggestion
		echo 'Wrote ' . number_format( $bytes ) . ' bytes to ' . $output_path . "\n";
```

ryelle: Or use `size_format`?

ryelle:
I tested out the script and it worked for me 👍🏻  

> Currently it treats w.org/main-test/ as the canonical source of content. Should that become w.org itself, or a staging site?
> Should it directly fetch the remote content (as it does now) or rely on 'setup:refresh' and use the local version of content?

If we use `main-test`/other staging site, we'll also need some process to push media to w.org & rewrite URLs. I think we could use w.org directly, but I don't know if there are other concerns. Syncing directly from remote -> pattern would be one less step where things could go sideways. None of these opinions are strongly-held though, if translation requires a different approach.

> Where should the URLs and corresponding output files be defined?

Maybe a json file, like this? For each object in the array, fetch `http://wordpress.org/canonical/wp-json/wp/v2/pages?context=wporg_export&slug=${slug}` , output pattern to `source/wp-content/themes/wporg-main-2022/patterns/${pattern}`, and you could also create template files at `source/wp-content/themes/wporg-main-2022/templates/${template}` for when we have new pages (I wouldn't overwrite those, though, since we might need to customize the header & footer variations).

```json
[
  {
    "slug": "front-page",
    "pattern": "front-page.php",
    "template": "front-page.html"
  },
  {
    "slug": "download",
    "pattern": "download.php",
    "template": "page-download.html"
  }
]
```

I’m not tied to JSON format, it’s just familiar :)

> How could it programatically discover newly created pages on the staging/canonical site?

I don’t think it should - IMO it’d be easy enough to add in new entries to a config file. Maybe an idea for the future though, once we’ve done enough pages that design/marketing are churning them out.

> Should this be done locally by devs, or as part of the sandbox sync?

Both? I would want to check it locally first, to make sure nothing unexpected changes. So to start, I think a local workflow is fine: dev spins up local env, runs `build:patterns`, confirms changes in git, and commits to trunk, then runs sync on sandbox.

env/export-content.php
tellyworth: Oh yeah I definitely a word out there :)

It's not loading WP code so size_format() doesn't exist.

tellyworth:
> If we use `main-test`/other staging site, we'll also need some process to push media to w.org & rewrite URLs. I think we could use w.org directly, but I don't know if there are other concerns. Syncing directly from remote -> pattern would be one less step where things could go sideways. None of these opinions are strongly-held though, if translation requires a different approach.

That's a great point about pushing media files. I know @dd32 had some concerns about subtle differences in behaviour that we might trigger by having the post content on the prod site itself. Not sure if it's easier to solve those problems, or solve the media problems. Let's explore ideas and see what seems like the most natural solution.

> Maybe a json file

Yeah that seems like a good place to start. I think I agree that defining the pages in code makes sense for now.

> > Should this be done locally by devs, or as part of the sandbox sync?
> 
> Both? I would want to check it locally first, to make sure nothing unexpected changes. So to start, I think a local workflow is fine: dev spins up local env, runs `build:patterns`, confirms changes in git, and commits to trunk, then runs sync on sandbox.

I think `build:patterns` should be easy to run on a sandbox since it's vanilla PHP that doesn't require a wp-env or anything. (It can't really run locally in the Docker environment since it needs to be able to write to the `/source/` files).

If it does run at sync time, would it need to push a commit to get the latest changes into Git? Or would we use svn as the source of truth for those generated pattern files?

One possibility that would be fairly easy given existing tooling is:

1. A dev runs `build:patterns` locally.
2. Open a PR with those content changes and merge.
3. Run the sync script on sandbox.
4. Sync script also runs `build:patterns` and checks for any diffs; if so, stop/warn/do something?

Or alternatively, steps 1-2 as above, then:

3. Run a separate `sync-content` script that only deploys changes from `wporg-main-2022/patterns/`.

Still thinking it over, I think it'll become more clear as we experiment.

tellyworth:
Pushed some changes to implement the json manifest idea.

tellyworth:
There's probably more to do here but I've un-drafted the PR.

@dd32 should I look at incorporating a first stab at translation in this PR, or better to merge this and move on to translation in a fresh PR?

dd32:
> should I look at incorporating a first stab at translation in this PR, or better to merge this and move on to translation in a fresh PR?

I would think second PR, although it depends on how translation is handled ultimately. If we can inject `<?php _e( ...)` into the pattern files, then that might be a good way to go. If we're going to generate locale pattern files though, I'm not convinced that this is a good place to store them (due to the long-process required to go from GlotPress -> live-in-production)

ryelle:
This script is working well for me with the manifest 👍🏻 

> If it does run at sync time, would it need to push a commit to get the latest changes into Git? Or would we use svn as the source of truth for those generated pattern files?

Would we have a separate repo for the patterns? if they're part of the theme, then i think they'd have to be committed to git (probably first, to preserve the Git -> SVN flow).

I think your first flow (1-4) matches what I expect. Step 4 might hit some issues if editors are updating content while we're trying to deploy, so maybe a `--force` flag or just warning on conflicts would be best.

tellyworth:
I'll merge this now and we can discuss the next pieces in #55.

PR #103, opened 2022-08-25
Preload the the heading fonts.

### LINKED ISSUE #92: Preload fonts to avoid initial layout shift
Ref: https://wordpress.slack.com/archives/C02QB8GMM/p1660634455683169?thread_ts=1658956086.933959&cid=C02QB8GMM

There's an initial layout shift because the fonts we use usually don't exist in users' locals,
the text would be rendered with local fonts first and then changed to network fonts after they are downloaded.

|  Before | After |
|--------|-------|
| image | image |

layout shift

dd32:
This uses the changes in https://github.com/WordPress/wporg-mu-plugins/pull/247 to preload the homepage heading fonts.

Fixes #92

### Screenshots

| Before | After |
|--------|-------|
| image  | image |

1. View source of https://wordpress.org/, note no preloaded fonts.
2. Apply https://github.com/WordPress/wporg-mu-plugins/pull/247 + This PR
3. View source of https://wordpress.org, note the preloaded fonts after the opening `<head>`

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -2,6 +2,8 @@
 
 namespace WordPressdotorg\Theme\Main_2022;
 
+use function WordPressdotorg\MU_Plugins\Global_Header_Footer\is_rosetta_site;
+
 require_once( __DIR__ . '/inc/page-meta-descriptions.php' );
 require_once( __DIR__ . '/inc/hreflang.php' );
 require_once( __DIR__ . '/inc/capabilities.php' );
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Same question about whether `is_callable` is necessary, but otherwise this is also good to merge once the mu PR is merged 👍🏻

source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -48,6 +50,11 @@ function enqueue_assets() {
 
 	// Preload the heading font(s).
 	if ( is_callable( 'global_fonts_preload' ) ) {
+		// TODO: Load required fonts for rosetta sites once they are supported.
+		if ( is_rosetta_site() ) {
+			return;
+		}
```
renintw, line 56: Although we're not supporting any rosetta sites yet, this check is still added in case one day we start to support and forget to update this part, making it load unnecessary extra font files.

source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -48,12 +50,15 @@ function enqueue_assets() {
 
 	// Preload the heading font(s).
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		// All headings.
-		global_fonts_preload( 'EB Garamond Latin' );
+		// TODO: Load required fonts for rosetta sites once they are supported.
```
outdoor2kode, line 67: Although we're not supporting any rosetta sites yet, this check is still added in case one day we start to support and forget to update this part, making it load unnecessary extra font files.

outdoor2kode: Issue opened: https://github.com/WordPress/wporg-main-2022/issues/136

PR #106, opened 2022-08-25
Blocks: Create a stub Download Modal "block"

ryelle:
This moves the download modal code into a subfolder in `src`, so that we can create additional blocks without needing a custom `webpack.config.js`. By default, `wp-scripts` will look in the src folder for any `block.json` files, and pull out the scripts referenced. We can use that behavior without creating a proper block by adding a `block.json` file with just the script. This "block" is not registered anywhere, and the JS is still enqueued the same way it was before.

Other blocks can be added as sibling folders to `download`, and they'll automatically be built too.

See #52 — eventually this will be a real block, so this won't be a workaround forever.

### Screenshots

No visual changes.

1. Build the project `yarn workspace wporg-main-2022-theme build`
2. It should generate new files inside `build/download`
3. Load the download page
4. The script should load
5. The modal should open on click


source/wp-content/themes/wporg-main-2022/src/components/download-modal/index.js
```diff
@@ -6,7 +6,7 @@ import { useEffect, useState } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import DownloadModal from './DownloadModal';
+import DownloadModal from './download-modal';
```
ryelle, line 9: Renamed this file to follow [the WP coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#naming): files should be lowercase, words separated with hypens.

adamwoodnz (approved):
:shipit: Works for me

I just added 1 commit to simplify the init code a little

PR #107, opened 2022-08-26
Add "Cycle Headings" block to animate through a list of headings.

### LINKED ISSUE #101: Homepage: Try a variety of headlines to cycle through
The homepage reads "WordPress: Publish your _passion_". Since WordPress is more than that, it would be good to see a range of options there in a rotating headline: “Publish your ______?” And we can get quirky there. Something like 3-5 good options, with the specific copy to be explored.

As a simple starting point, the headline block does not need to be interactive, and can simply surface hardcoded options:

mock

This would presumably simplify ensuring a well translatable and accessible solution, and can grow into bigger things if need be in the future.

Semantically, one option could be to have the first heading be static, the subsequent 2-5 headings to swap out with be fake and, then pair it with a hidden heading announcing the a full range. Something like:

* “Publish your passion”
* “Publish your diary”
* […]

And then in the hidden heading, “Publish your passion, your diary, your […], or your […] with WordPress”. Something to explore and vet in practice.

ryelle:
Add a new block, "Cycle Headings." This block renders "WordPress: _____", where the blank space cycles through a set of phrases.

- Publish your passion
- Grow your business
- Flex your freedom.

These change after 5 seconds.

Semantically, it renders a hidden heading (h1) & paragraphs, and this visual content is hidden from screen readers. JS is used to swap the displayed text, with a CSS transition on the opacity. This a version of the code used on [the p2 landing site](https://wordpress.com/p2/), tweaked to work with the block output.

The front-page template is updated to use this block.

Fixes #101.

Screen reader users will hear "Heading level 1: WordPress: Publish your passion. Publish your passion. Grow your business. Flex your freedom."

There is no change to the headings if you're using "prefers reduced motion", just to be safe.

### Screenshots

https://user-images.githubusercontent.com/541093/186962227-4ffb1c4c-e38f-4cd1-b50d-fcc6c47ce58a.mp4

In the editor, it's just a static block with the main heading. Eventually we could offer some way to enter a few headings & pass them through as attributes (TBH I could add this pretty quickly now, just as some text inputs in the sidebar). Font size & spacing can be controlled here, though it defaults to the H1 styling.

Screen Shot 2022-08-26 at 1 42 30 PM

_Fonts don't load in the editor on local envs, it's correct on w.org_

1. Build the branch, `yarn workspace wporg-main-2022-theme build`
2. Load the front page
3. Wait 5 seconds, you should see the heading change
4. It should cycle through all 3 headings

Optionally…
- Try with a screen reader - it should make sense
- Try with reduced motion on (will need to reload the page) - it won't change

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
@@ -4,7 +4,7 @@ const MS_DELAY_BETWEEN_ITEMS = 500;
 
 function textFragmentIterator() {
 	let currentIteration = 0;
-	const fragments = Array.from( document.querySelectorAll( '.wp-block-wporg-cycle-headings p > span' ) ).map(
+	const fragments = Array.from( document.querySelectorAll( '.wp-block-wporg-cycle-headings p' ) ).map(
 		( elem ) => elem.innerHTML
 	);
 
--- a/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
@@ -1,6 +1,6 @@
 // Set up the dynamic text in the hero section of the Home page.
-const MS_DURATION_PER_ITEM = 5000;
-const MS_DELAY_BETWEEN_ITEMS = 500; // Should equal the sum of transition duration in CSS.
+const DURATION_PER_ITEM_MS = 5000;
+const DELAY_BETWEEN_ITEMS_MS = 500; // Should equal the sum of transition duration in CSS.
 
 function textFragmentIterator() {
 	let currentIteration = 0;
@@ -20,7 +20,7 @@ function init() {
 
 	window.setTimeout( () => {
 		targetNode.classList.add( 'is-ready' );
-	}, MS_DURATION_PER_ITEM - 1 );
+	}, DURATION_PER_ITEM_MS - 1 );
 
 	window.setInterval( () => {
 		targetNode.classList.remove( 'fade-in' );
@@ -28,8 +28,8 @@ function init() {
 		window.setTimeout( () => {
 			targetNode.innerHTML = getNextTextFragment();
 			targetNode.classList.add( 'fade-in' );
-		}, MS_DELAY_BETWEEN_ITEMS );
-	}, MS_DURATION_PER_ITEM );
+		}, DELAY_BETWEEN_ITEMS_MS );
+	}, DURATION_PER_ITEM_MS );
 }
 
 const wantMotionQuery = window.matchMedia( '(prefers-reduced-motion)' );
--- a/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
@@ -4,8 +4,8 @@ const MS_DELAY_BETWEEN_ITEMS = 500;
 
 function textFragmentIterator() {
 	let currentIteration = 0;
-	const fragments = Array.from( document.querySelectorAll( '.wp-block-wporg-cycle-heading p' ) ).map(
-		( elem ) => elem.textContent
+	const fragments = Array.from( document.querySelectorAll( '.wp-block-wporg-cycle-heading p > span' ) ).map(
+		( elem ) => elem.innerHTML
 	);
 
 	return () => {
@@ -25,7 +25,7 @@ window.setInterval( () => {
 	targetNode.classList.remove( 'fade-in' );
 
 	window.setTimeout( () => {
-		targetNode.textContent = getNextTextFragment();
+		targetNode.innerHTML = getNextTextFragment();
 		targetNode.classList.add( 'fade-in' );
 	}, MS_DELAY_BETWEEN_ITEMS );
 }, MS_DURATION_PER_ITEM );
--- a/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
@@ -1,3 +1,6 @@
+// Code originally forked from https://wordpress.com/p2/
+// https://github.com/Automattic/p2/blob/master/themes/marketing/src/js/hero-dynamic-text.js
+
 // Set up the dynamic text in the hero section of the Home page.
 const DURATION_PER_ITEM_MS = 5000;
 const DELAY_BETWEEN_ITEMS_MS = 500; // Should equal the sum of transition duration in CSS.
--- a/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
@@ -1,5 +1,5 @@
 // Set up the dynamic text in the hero section of the Home page.
-const MS_DURATION_PER_ITEM = 4000;
+const MS_DURATION_PER_ITEM = 5000;
 const MS_DELAY_BETWEEN_ITEMS = 500;
 
 function textFragmentIterator() {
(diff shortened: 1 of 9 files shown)


source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
```diff
@@ -0,0 +1,39 @@
+// Set up the dynamic text in the hero section of the Home page.
+const MS_DURATION_PER_ITEM = 5000;
+const MS_DELAY_BETWEEN_ITEMS = 500; // Should equal the sum of transition duration in CSS.
+
+function textFragmentIterator() {
+	let currentIteration = 0;
+	const fragments = Array.from( document.querySelectorAll( '.wp-block-wporg-cycle-headings p' ) ).map(
```
StevenDufresne, line 10: Expecting the markup to be `p` tag could be fragile (ie: the whole `h1` could be replaced). Does it make sense to just use a class on the fragments?

source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
```diff
@@ -0,0 +1,39 @@
+// Set up the dynamic text in the hero section of the Home page.
+const MS_DURATION_PER_ITEM = 5000;
+const MS_DELAY_BETWEEN_ITEMS = 500; // Should equal the sum of transition duration in CSS.
```
StevenDufresne: I'm used to seeing the type appended at the end: `DELAY_BETWEEN_ITEMS_MS`. It appears like Gutenberg also follows that convention. Should we align with Gutenberg?

StevenDufresne (commented):
This is working as expected, although adding a background color in the editor adds a whole bunch of padding. 🤔 

Video: https://d.pr/i/scNJ24

StevenDufresne:
Also, If we were really concerned with performance, I think we could probably do this with CSS, but I'm fine with the simplicity of this solution. 👍  

I've added a couple of comments but they aren't blockers.

StevenDufresne (approved):

jasmussen:
Thanks for the PR and thanks for the expressive video (I haven't yet set up my local env to test, I'll be sure to do that for the future). I'd love a quick sanity check from @critterverse.

The animation speed for the fade seems good, but could potentially be faster (something to feel out and look at separately).

It's a little unfortunate that "Flex your Freedom" wraps after "your", when the previous two headlines wrap have "your passion" and "your business" on a new line. Can we do something with `<br />` or `&nbsp;` so that in more cases, the three phrases will wrap in the same way, with `your` as the first word on line two? I realize the challenge given it's a responsive design, there may not be a ton we can do, but if we had a great idea here, it would be good.

In general, the trio of headlines feels like it better captures the diversity of WordPress than the previous one. In that light, it seems good to keep moving forward 👍

critterverse:
So cool, this looks great!! 

>It's a little unfortunate that "Flex your Freedom" wraps after "your", when the previous two headlines wrap have "your passion" and "your business" on a new line.

+1. This is currently in a columns block with an empty column to the right to help with line width. I think we could get away with tweaking the column width on the left to be just a tiny bit smaller so that the "your" bumps down to the next line. What do you think @ryelle?

>I haven't yet set up my local env to test, I'll be sure to do that for the future

I also haven't done this yet since we were using a staging site up to this point. Is this something we definitely need to use Docker for? Wondering if there's an easier way to review 😅

source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
ryelle: I'm not sure I follow, what's fragile? We're controlling the rendered HTML output, so we know those `p`s will be there.

ryelle:
> This is working as expected, although adding a background color in the editor adds a whole bunch of padding. 🤔

Yeah, but that's the case for any heading/paragraph, not an issue in this PR. The padding there comes from the core GB block styles.

> If we were really concerned with performance, I think we could probably do this with CSS

The JS adds no dependencies, so I don't think we need to worry about it. If we get reports about it later we can look into that.

> Can we do something with `<br />` or `&nbsp;`

These both end up a little awkward at smaller screen sizes.

The `&nbsp;` causes a word break:

nbsp

And the `br` puts each word on its own line once it needs to wrap:

br

Adjusting the columns seems to work best for the large screen sizes, without the above issues 👍🏻 

Screen Shot 2022-08-29 at 11 38 27 AM

There is an issue at around 800-1000px wide where the "Publish…" heading wraps to 3 lines, but the rest stay on 2. At 800px, all three headings wrap to 3 lines so we don't have this jump anymore. It kicks in again at 430px wide, though. (video sped up to show the jumps)

https://user-images.githubusercontent.com/541093/187236688-109c44d5-1bec-4327-bafc-e07801b269db.mp4

> I also haven't done this yet since we were using a staging site up to this point. Is this something we definitely need to use Docker for? 

The local env that we've set up does use wp-env/docker, and we've created the setup instructions & scripts with the expectation that you're using that, but maybe something like Local could work if you can map the wp-content directory to `wp-content` in this repo? You'll still need to run `yarn setup:tools` to get all the dependencies though.

critterverse:
I wonder if it would work better to split the rotating part of the headline on its own line. I _think_ it would help with consistency in the number of lines of text as the headline is rotating:

headine-update

ryelle:
I tried that out and it didn't help - "Publish your passion" still wraps to a new line before Flex & Grow do.

critterverse:
I added a new page in [the Figma](https://www.figma.com/file/WHhRwVc4swYnPKdyVWj3du/Home-%26-Download-Pages?node-id=1945%3A19840) so we can play around with some options (cc @javierarce). Personally, I think I'd lean more towards a static headline if we can't find a way to avoid making things jump around on the page — this is especiallly going to be tricky across languages.

^ One different approach could be that the headline swaps each time you load the page — so each time you visit the landing, you'd see a different headline 🤔

ryelle:
We could try what the p2 site does, and fix the container height to the tallest option

| 2 lines | 3 lines |
|-----|-----|
| Screen Shot 2022-08-29 at 1 16 08 PM | Screen Shot 2022-08-29 at 1 16 05 PM |

But TBH I think this looks awkward with our large heading size.

Screen Shot 2022-08-29 at 1 18 28 PM

critterverse:
I looked at this too! But it agree it feels kind of awkward and pushes that CTA button down pretty far which is not ideal I suppose.

source/wp-content/themes/wporg-main-2022/src/cycle-headings/view.js
StevenDufresne: It relies on a specific markup tag, controlled or not. Future developers have to update code in 2 places when markup changes. If we were to use classes, that would not be the case.... and it's probably 1ms slower :) 

Aside, I wonder how much slower and whether the browser's bottom-up parsing of CSS would impact a page with a lot of `<p>` tags. 🤔

ryelle: Classes can change too, so I don't really see a difference. In any case, I'm going to leave this as is.

> I wonder how much slower and whether the browser's bottom-up parsing of CSS would impact a page with a lot of `<p>` tags

According to [this article](https://www.sitepoint.com/optimizing-css-id-selectors-and-other-myths/), if the page has 50,000 `p`, a whole 4ms. The issue comes more from long/complex relational selectors, but even so, we're better off focusing elsewhere for performance gains.

ryelle:
@critterverse Should I wait on a new design, or did you want me to try a random heading per page load?

critterverse:
I think we should give the random heading per page load a try! Cc @jasmussen in case you have strong feelings about keeping the cycle functionality.

jasmussen:
Happy to keep moving on this one!

ryelle:
I'm going to close this PR and open a new one for static random headings.

PR #108, opened 2022-08-29
Allow post content edits to be viewed by Editors

tellyworth:
Still WIP, but this adds another piece of the editing workflow: allow Editors to see post_content when viewing the site, overriding the pattern file. That way they can preview and "publish" their edits, and reviewers with the appropriate creds can see it.

This could be used in staging or production, depending on how confident we are and what makes sense for the workflow.

Still a few things to resolve:

* Make this work cleanly and sensibly for posts that use the old theme;
* Add a better indicator of post content vs pattern content than 'THIS IS POST CONTENT'
* Figure out permissions etc so that preview links can be sent to reviewers (if that even makes sense)

To test:

1. Run `yarn setup:refresh` to make sure you have some post content.
2. View `localhost:8888` while logged out and confirm it matches main-test content.
3. Log in and edit the front page post content with an obvious change. Click Update.
4. View `localhost:8888` while logged *in* and confirm you can see your change. Also `THIS IS POST CONTENT` at the top.
5. View `localhost:8888` while logged *out* and confirm you see the original pattern without your change.

Logged in view example with edit (red heading):

Screen Shot 2022-08-29 at 5 24 57 pm

Logged out view showing the pattern file instead of edited post_content:
Screen Shot 2022-08-29 at 5 24 44 pm

This would look better on a prod server because of missing media files.

See #55.

### DIFF
--- a/source/wp-content/mu-plugins/theme-switcher.php
+++ b/source/wp-content/mu-plugins/theme-switcher.php
@@ -90,6 +90,8 @@ add_action(
 						);
 
 						if ( $count > 0 ) {
+							$_wp_current_template_content = str_replace( [ '"layout":{"inherit":true},"className":"entry-content",', ' entry-content' ], '', $_wp_current_template_content );
+
 							add_action( 'admin_bar_menu', __NAMESPACE__ . '\admin_bar_preview_indicator', 1000 );
 						}
 					}
--- a/source/wp-content/mu-plugins/theme-switcher.php
+++ b/source/wp-content/mu-plugins/theme-switcher.php
@@ -56,6 +56,7 @@ function admin_bar_preview_indicator( $wp_admin_bar ) {
 		[
 			'id'    => 'preview-indicator',
 			'title' => '<span class="ab-icon dashicons-admin-appearance"></span> ' . $text,
+			'href'  => get_permalink( get_queried_object_id() ),
 			'meta'  => [
 				'class' => 'preview-indicator',
 			],
--- a/source/wp-content/mu-plugins/theme-switcher.php
+++ b/source/wp-content/mu-plugins/theme-switcher.php
@@ -15,6 +15,11 @@ namespace WordPressdotorg\MU_Plugins\Theme_Switcher;
  * Helper to check the requested page against our new page list.
  */
 function should_use_new_theme() {
+	// Front-page sites only.
+	if ( function_exists( '\get_blog_details' ) && '/' !== \get_blog_details( null, false )->path ) {
+		return false;
+	}
+
 	// Request to resolve a template.
 	if ( isset( $_GET['_wp-find-template'] ) ) {
 		return true;
--- a/source/wp-content/mu-plugins/theme-switcher.php
+++ b/source/wp-content/mu-plugins/theme-switcher.php
@@ -52,10 +52,6 @@ if ( 'production' !== wp_get_environment_type() ) {
 // Only if the user is logged in and can edit posts:
 // Override the block template, to force loading post_content instead of the hard-coded pattern file.
 // This is so that content and design can be edited or created in-place.
-// TODO:
-// * Make this work cleanly and sensibly for posts that use the old theme;
-// * Add a better indicator of post content vs pattern content than 'THIS IS POST CONTENT'
-// * Figure out permissions etc so that preview links can be sent to reviewers
 add_action(
 	'init',
 	function() {
--- a/source/wp-content/mu-plugins/theme-switcher.php
+++ b/source/wp-content/mu-plugins/theme-switcher.php
@@ -25,6 +25,11 @@ function should_use_new_theme() {
 		return true;
 	}
 
+	// Preview. Can't call is_preview() this early in the process.
+	if ( isset( $_GET['preview'] ) && isset( $_GET['preview_id'] ) ) {
+		return true;
+	}
+
 	$new_theme_pages = array(
 		'/',
 		'/download/',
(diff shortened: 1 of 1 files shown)


tellyworth:
Updated to only show post content in preview mode, which neatly solves those TODO issues. Updated screenshot in preview mode:

Screen Shot 2022-08-30 at 5 30 52 pm

Updated test plan is:


1. Run `yarn setup:refresh` to make sure you have some post content.
2. View `localhost:8888` while logged out and confirm it matches main-test content.
3. Log in and edit the front page post content with an obvious change. Click Preview.
4. View the preview in a new tab while logged in and confirm you can see your change. Also `Previewing post content` in the admin bar.
5. View localhost:8888 while logged out and confirm you see the original pattern without your change.


There are still two problems to address:

1. The layout in preview mode is incorrect. I think something is wrapping the post content in an extra div, haven't figured out where it's coming from yet.
2. Preventing editors from creating new pages. I tried removing the `publish_pages` cap but that changes the Publish button to Submit for Review and results in the post status changing to Draft, which is not good. `create_pages` seems promising but I'm yet to get it to work.

tellyworth:
@ryelle I think the preview layout bug is the only blocker to testing this in production with design folks, unless you see other problems?

source/wp-content/mu-plugins/theme-switcher.php
```diff
@@ -43,3 +48,54 @@ function should_use_new_theme() {
...
+						$_wp_current_template_content = preg_replace(
+							'#<!-- wp:pattern {"slug":"wporg-main-2022/[\w-]+"} /-->#',
+							'<!-- wp:post-content {"layout":{"inherit":true},"style":{"spacing":{"blockGap":"0px"}}} /-->',
+							$_wp_current_template_content,
+							1,
+							$count
+						);
+
+						if ( $count > 0 ) {
+							add_action(
```
ryelle: > The layout in preview mode is incorrect. I think something is wrapping the post content in an extra div, haven't figured out where it's coming from yet.

It's from the group — the `layout` should only be on the `post-content` block (see `templates/page.html`), having it on the group says that post-content should be constrained.

```diff
- <!-- wp:group {"tagName":"main","layout":{"inherit":true},"className":"entry-content","style":{"spacing":{"blockGap":"0px"}}} -->
- <main class="wp-block-group entry-content">
+ <!-- wp:group {"tagName":"main"} -->
+ <main class="wp-block-group">
```

(the style can stay, it's also duplicated on the post-content but has no effect)

You could add this plain `str_replace` here (so it only replaces if we've already found the pattern), or update the preg_replace above (but I didn't want to mess with regex right now 🙂 ). In any case, this fixes the layout bug.

```suggestion
							$_wp_current_template_content = str_replace( [ '"layout":{"inherit":true},"className":"entry-content",', ' entry-content' ], '', $_wp_current_template_content );
							add_action(
```

source/wp-content/mu-plugins/theme-switcher.php
```diff
@@ -43,3 +48,54 @@ function should_use_new_theme() {
...
+								'admin_bar_menu',
+								function( $wp_admin_bar ) {
+									$text = 'Previewing post content';
```
ryelle: What do you think about making this a link to the current live page?

source/wp-content/mu-plugins/theme-switcher.php
```diff
@@ -43,3 +48,54 @@ function should_use_new_theme() {
...
+// Only if the user is logged in and can edit posts:
+// Override the block template, to force loading post_content instead of the hard-coded pattern file.
+// This is so that content and design can be edited or created in-place.
+add_action(
+	'init',
+	function() {
+		if ( current_user_can( 'edit_posts' ) ) {
+			add_filter(
+				'template_include',
+				function( $template ) {
```
ryelle: I don't feel strongly about this, but having these nested anon functions could be a little hard to understand at quick glance, maybe this one could be named?

ryelle (approved):
Works for me — I tried editing the front page, downloads page, and another not-yet-redesigned page and all three previewed using the new theme 👍🏻 

Left a comment about fixing the layout bug, the other comments are not blockers.

source/wp-content/mu-plugins/theme-switcher.php
tellyworth: Good idea, done.

source/wp-content/mu-plugins/theme-switcher.php
tellyworth: Thanks! Yeah the str_replace seems as good a solution as any for now.

source/wp-content/mu-plugins/theme-switcher.php
tellyworth: Agreed, I moved the inner filter and action to named functions.

tellyworth:
I pushed some more changes today after testing this in a sandbox. My plan is:

* Commit `theme-switcher.php` to prod in `mu-plugins/main-network`
* Remove the call in `mu-plugin/main-network/site-main.php` that loads `mu-plugins/rosetta-network/2022-theme-switcher.php`, thus replacing `2022-theme-switcher.php` with this updated version of `theme-switcher.php`.

In other words, this version of `theme-switcher.php` will take over as the theme switcher in production.

I've tested that on my sandbox and it seems to work solidly now, including for a few edge cases like `w.org/download/?junk` which failed before.

Once that's deployed and settled, I'll add some design folks as Editors to the main site and document the workflow for them. Key points to cover:
* Preview will show post content; saving edited post content won't affect the live site.
* They should double check the live page just in case.
* Don't edit page metadata, featured image, etc - that will affect the live page.
* Don't trash a page or convert it to a draft! That will affect the live page.
* Be careful creating new pages - if you publish one it will be live.

I'll do that tomorrow my time unless a blocker turns up.

Note that the build and sync workflow doesn't yet include anything for deploying `theme-switcher.php` changes to production. We'll need to solve that going forward too, that can be a future issue/PR though.

source/wp-content/mu-plugins/theme-switcher.php
```diff
@@ -25,21 +30,96 @@ function should_use_new_theme() {
...
+	// If the new theme is the active theme, we should use it, otherwise, we should use the old theme.
+	return ( 'wporg-main-2022' === get_option( 'stylesheet' ) );
 }
 
-// Always show admin bar.
-add_filter( 'show_admin_bar', '__return_true' );
+// Always show admin bar on local test site
+if ( 'local' === wp_get_environment_type() ) {
+	add_filter( 'show_admin_bar', '__return_true' );
+}
```
ryelle: I think we can remove this entirely now that the admin bar PR merged.

PR #111, opened 2022-08-31
Theme: Add workaround for unsupported CSS4 selectors

### LINKED ISSUE #78: Content is not centered when browser doesn't support CSS4 selectors.
**Describe the bug**
Similar to issue we saw with the navigation, the `group` block also uses `:where` and `:not`.

**For Example:**
When this is missing, the content is not centered.
```
.wp-block-group.wp-container-46 > :where(:not(.alignleft):not(.alignright):not(.alignfull))
```

### Example
**OS**: Catalina
**Safari**: 13.1

Screen Shot 2022-08-14 at 7 32 54 PM

iandunn:
Some Gutenberg blocks use CSS4 selectors, which don't work on some older browsers that should still be supported. This adds some temporary styles to work around that while the upstream bug is being fixed.

Fixes #78

### Screenshots

* [before](https://user-images.githubusercontent.com/484068/187569996-a5bbdd48-7563-49ee-bea5-e06e75c5a7fb.png)
* [after](https://user-images.githubusercontent.com/484068/187569895-5cb11bf4-4ce4-43fb-ba5a-73bcef3cad38.png)

### Testing

Compare the Homepage and Download page in an older browser, like [Opera 70](https://get.opera.com/ftp/pub/opera/desktop/70.0.3728.95/), against a modern browser. You may want to [disable autoupdates](https://superuser.com/questions/828009/how-to-disable-automatic-updates-in-opera) so it remains at an older version.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/style.css
+++ b/source/wp-content/themes/wporg-main-2022/style.css
@@ -227,3 +227,34 @@
 	padding-left: 0;
 	margin-bottom: 0;
 }
+
+
+/*
+ * Workaround for Gutenberg CSS4 selectors on older browsers.
+ *
+ * @link https://github.com/WordPress/gutenberg/issues/43723.
+ *
+ * Reproducing gutenberg_get_layout_style() isn't practical, so this just hardcodes the needed values.
+ *
+ * Remove this when the above issue is fixed in a stable release.
+ */
+.wp-block-group[class*="wp-container-"],
+.wp-block-columns[class*="wp-container-"],
+.wp-block-group[class*="wp-container-"] > *,
+.wp-block-columns[class*="wp-container-"] > *,
+[class*="wp-container-"] .wp-block-buttons {
+	margin-left: auto !important;
+	margin-right: auto !important;
+}
+
+.wp-block-group[class*="wp-container-"].entry-content,
+.wp-block-group[class*="wp-container-"][class*="is-content-justification-"] > * {
+	margin-left: unset !important;
+	margin-right: unset !important;
+}
+
+#benefits .wp-block-group[class*="wp-container-"],
+#community .wp-block-group[class*="wp-container-"],
+#editor .wp-block-group[class*="wp-container-"] {
+	max-width: 520px;
+}
--- a/source/wp-content/themes/wporg-main-2022/style.css
+++ b/source/wp-content/themes/wporg-main-2022/style.css
@@ -238,23 +238,18 @@
  *
  * Remove this when the above issue is fixed in a stable release.
  */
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-main-2022/style.css
```diff
@@ -227,3 +227,34 @@
...
+/*
+ * Workaround for Gutenberg CSS4 selectors on older browsers.
+ *
+ * @link https://github.com/WordPress/gutenberg/issues/43723.
+ *
+ * Reproducing gutenberg_get_layout_style() isn't practical, so this just hardcodes the needed values.
+ *
+ * Remove this when the above issue is fixed in a stable release.
+ */
+.wp-block-group[class*="wp-container-"],
```
ryelle: These container classes aren't guaranteed to be there - relying on them is why we needed [this fix for GB 13.7](https://github.com/WordPress/wporg-news-2021/pull/387). The new class that will be used for this type of layout is `.is-layout-constrained`, so we could repeat the core styles like so:

```
body .is-layout-constrained > * {
	max-width: var(--wp--style--global--content-size);
	margin-left: auto !important;
	margin-right: auto !important;
}

body .is-layout-constrained > .alignwide {
	max-width: var(--wp--style--global--wide-size);
}

body .is-layout-constrained > .alignfull {
	max-width: unset;
	margin-left: unset !important;
	margin-right: unset !important;
}
```

That seems to work without relying on the container class.

source/wp-content/themes/wporg-main-2022/style.css
```diff
@@ -227,3 +227,34 @@
...
+}
+
+.wp-block-group[class*="wp-container-"].entry-content,
+.wp-block-group[class*="wp-container-"][class*="is-content-justification-"] > * {
+	margin-left: unset !important;
+	margin-right: unset !important;
+}
+
+#benefits .wp-block-group[class*="wp-container-"],
+#community .wp-block-group[class*="wp-container-"],
```
ryelle: Why are these needed? I don't think they apply to anything, unless there's a template change I'm missing.

source/wp-content/themes/wporg-main-2022/style.css
iandunn: When I opened the PR the latest Gutenberg didn't have the constrained layout. That code works now that 14 is out though 👍🏻

source/wp-content/themes/wporg-main-2022/style.css
iandunn: They were needed in 13.9, but your above code works with 14 👍🏻

PR #113, opened 2022-08-31
Block: Add "Random Headings" block

### LINKED ISSUE #101: Homepage: Try a variety of headlines to cycle through

ryelle:
Fixes #101, See #107. This adds a new block, "Random Headings," which renders a random heading from our set of 3 on every page load.

Since these just render static headings, we don't have the accessibility concerns of #107.

I did copy over the column width changes so that at the default desktop size (>1000), the heading wraps before "your".

### Screenshots

| Publish | Grow | Flex |
|---------|------|------|
| publish-desktop | grow-desktop | flex-desktop |
| publish-mobile | grow-mobile | flex-mobile |

1. Build this branch
2. Open the font page of the site — the block has been replaced in the pattern
3. Reload a few times, the heading should change


critterverse:
Thanks @ryelle, this looks good to me! 🚢

StevenDufresne (approved):
Yep, this looks good. 

It could equally be called `random-heading` since the output is only ever 1 heading... but that's just opinion.

ryelle:
> It could equally be called random-heading since the output is only ever 1 heading... but that's just opinion.

well now that you've said that… 😆  I copied the structure from #107 & left it plural. Updated in https://github.com/WordPress/wporg-main-2022/pull/113/commits/7a54a005ace48b3595210f786245a3e6bb8dd0ac.

PR #115, opened 2022-09-02
Only revert to old theme on sites with `wporg-main-2022` active

tellyworth:
The old logic was backwards, it forced the main theme active on other sites. That triggered some fatals on the forums requiring a revert.


tellyworth:
This works well on my sandbox, merging.

PR #116, opened 2022-09-06
Fix export context bug

tellyworth:
The recent change to `theme-switcher.php` in production broke the export context, so `https://wordpress.org/main-test/wp-json/wp/v2/pages?context=wporg_export` no longer works.

This fixes it by ensuring the correct filters are added.

1. Copy `theme-switcher.php` to your sandbox.
2. Confirm that `https://wordpress.org/main-test/wp-json/wp/v2/pages?context=wporg_export` does not trigger an error when sandboxed.

Note that I've also changed `setup:refresh` to fetch content from production rather than main-test, since the test site is no longer in use. It's impossible to test that script against a sandbox though, since Docker will always hit production servers.

### DIFF
--- a/env/refresh.sh
+++ b/env/refresh.sh
@@ -2,5 +2,5 @@
 
 # Refresh the content from the server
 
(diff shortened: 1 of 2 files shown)


ryelle (approved):
Looks good, and I'm going to merge this so I can use it in another PR 🙂

PR #118, opened 2022-09-06
Try parsing block content and replacing strings with i18n functions

ryelle:
See #15 — This updates the `build:patterns` script to inject i18n functions into the patterns, which can then be synced into GlotPress using the regular code translation process.

This uses the parsers from the pattern directory project to extract strings from the post/pattern content, then wraps them in `<?php _e( '%s', 'wporg' ); ?>`. The parsers themselves have not been updated, but the main parser class (`BlockParser` here, `parser.php`) has been updated to add the `replace_with_i18n` function. I also had to tweak `/parsers/Button.php` to support having a shortcode in `href`.

It would probably be better if these files were not duplicated, at least the `parsers/*`, but I'm not sure what the best approach would be.

The `export-content` script has been updated so it now needs to be called via wp-cli, so that we can use the block parsing & serialization functions. `build:patterns` handles that behind the scenes. The only configuration allowed now is the manifest file.

### Screenshots

There should be no visual change.

1. Run `yarn build:patterns`
2. The pattern files should not output text directly, they should use PHP i18n functions.


env/export-content/includes/utils.php
```diff
@@ -0,0 +1,74 @@
...
+
+namespace WordPress_org\Main_2022\ExportToPatterns;
+
+require __DIR__ . '/parser.php';
+
+/**
+ * Generate the pattern content from a URL.
+ */
+function generate_pattern( $url, $output_path ) {
+	$data = file_get_contents( $url );
```
tellyworth, line 12: This works fine against production, but fails when sandboxed.

I added some code taken from `import-content.php` to replace the `file_get_contents()` call with `wp_remote_get()` and use `filter_curl_options` to bypass the sandbox, which seemed to do the trick.

That's not a blocker for this PR. If you want to merge as-is I'll open a fresh PR with the sandbox fix - didn't want to push it here because you might have a reason to not want to bypass the SB.

tellyworth (approved):
Looks good to me, other than the sandbox issue which isn't a blocker. I haven't reviewed the parser code in detail.

env/export-content/includes/utils.php
ryelle: Ah, that wasn't intentional - I didn't look too deeply into what the `curl` code was doing in `import-content.php`.

> If you want to merge as-is I'll open a fresh PR with the sandbox fix

I'd appreciate that 🙂

PR #122, opened 2022-09-13
Add/use latest news

StevenDufresne:
Update front-page template to apply changes made in `wporg-mu-plugins` block.

Needs to be deployed at the same time as: https://github.com/WordPress/wporg-mu-plugins/pull/267.


ryelle (approved):
I made one change, since the new block doesn't save any content this was triggering a validation error when the pattern loaded in the editor.

PR #125, opened 2022-09-14
Add RTLCSS support

### LINKED ISSUE #124: Build RTL CSS
Check if we need to build RTL CSS from our vanilla `style.css` or block styles.

Could use a custom `postcss` config? or a custom webpack config? See https://github.com/WordPress/gutenberg/issues/29605.

See #15

ryelle:
Fixes #124. We need to be able to run `rtlcss` on the blocks (so, during `wp-scripts build` & `wp-scripts start`), and on the CSS in `style.css`. Ideally, these would be the same process. We can add an `rtlcss` step to wp-scripts with a custom webpack config.

Instead of using the regular theme's `style.css`, I've moved the CSS into the `src` directory as another "fake block" (like the download modal). This way, we can use the same build step for everything. This is admittedly a kind-of hacky workaround, and it requires an empty JS file that isn't used anywhere.

But it works — with the added webpack config, all the CSS files also generate RTL versions.

### Screenshots

There aren't many CSS differences to point to, but you can see the difference in the features list and in the download modal.

| Before | After |
|--------|-------|
| before-css | after-css |
| before-modal | after-modal |

1. Run `yarn workspace wporg-main-2022-theme build`
2. You should see new files in `build`, a new folder called `styles` & new CSS files in the `download` folder.


StevenDufresne (approved):
I'm okay with this. I agree the `style` block is weird, but I can't think of a better approach without digging really deep which I don't think we need to do yet.

PR #126, opened 2022-09-15
Update inline styles & class names when using RTL display

### LINKED ISSUE #123: Fix inline styles when using an RTL display
See #15.

When using an RTL language, most of the alignments work out (columns is smart about it, for one 🎉). But there are places where spacing is added to the left/right and this is not switched when the language is changed. See "Powerful and empowering" heading — this is a `margin-right` that should be `margin-left` on RTL sites.

Other blocks explicitly set the text alignment, which is incorrect when the text direction flips. See the single line in the very first section — this has `wp:paragraph {"align":"left"…` and `class="has-text-align-left…`, which should be right (or removed, it's not actually necessary, but it could be in other places).

ryelle:
Fixes #123 — Flip the left/right spacing & text alignment classes when using an RTL display. In some places, we've added padding or margin to a section, which when flipped, breaks the visual alignment. The padding/margin should also get flipped. Likewise, text alignment should be flipped (although in most cases the alignment has [been removed](https://github.com/WordPress/wporg-main-2022/commit/682017ddb362c17605ba5f0dec1482c56005f1e7)).

When combined with #125, there shouldn't be any more visual bugs in RTL.

### Screenshots

| LTR | RTL Before | RTL After |
|-----|------------|-----------|
| ltr-php-benefits | before-php-benefits | after-php-benefits |
| ltr-php-download | before-php-download | after-php-download |
| ltr-php-latest-news | before-php-latest-news | after-php-latest-news |

1. Set your site language to an RTL language (`he_IL` works, or Arabic) in Settings, the locale files will download and your site will flip
2. View the homepage, download page
3. The content is flipped, but everything should still look aligned at the edges (for example, the "Powerful and empowering" heading)

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -15,6 +15,7 @@ require_once( __DIR__ . '/src/random-heading/index.php' );
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
 add_filter( 'render_block_core/pattern', __NAMESPACE__ . '\prevent_arrow_emoji', 20 );
+add_filter( 'render_block_core/pattern', __NAMESPACE__ . '\convert_inline_style_to_rtl', 20 );
 add_filter( 'wp_img_tag_add_loading_attr', __NAMESPACE__ . '\override_lazy_loading', 10, 2 );
 
 /**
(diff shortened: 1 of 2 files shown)


adamwoodnz:
Looks good to me. So strange having the punctuation at the other end of the line as a ltr native!

Any idea if these arrows should still point to the right?

Screen Shot 2022-09-20 at 10 43 38 AM

ryelle:
> Any idea if these arrows should still point to the right?

Yeah it looks awkward, but it's part of the string itself. When it's translated, the RTL translation should include the reversed arrow.

adamwoodnz (approved):
One question above, but not a blocker

PR #131, opened 2022-10-05
Add & sync download subpages

ryelle:
Fixes #129 — This branch has all the download subpages in one place. This contains #132 & #134, and uses https://github.com/WordPress/wporg-parent-2021/pull/48. 

This enables the syncing for all 4 download subpages: [Releases](https://wordpress.org/download/releases/), [Nightly](https://wordpress.org/download/beta-nightly/), [Counter](https://wordpress.org/download/counter/), [Source](https://wordpress.org/download/source/). Once we merge this, the new design can be enabled on these pages.

There was discussion about the subnavigation bar on #129 — I've added a basic iteration but the actual alignment can be iterated on in the editor after merge, if necessary. 

### Screenshots

| Beta/Nightly | Counter | Releases | Source |
|--------------|---------|----------|--------|
| beta-desktop | counter-desktop | releases-desktop | source-desktop |
| beta-mobile | counter-mobile | releases-mobile | source-mobile |

Note, the above screenshots were taken before the mobile nav button was updated to use the down-chevron - it looks like this now:
Screen Shot 2022-10-14 at 4 55 38 PM

<table>
<thead>
<tr>
<th colspan="2">Mobile "dropdown" style nav open</th>
</tr>
</thead>
<tbody>
<tr>
<td><a target="_blank" rel="noopener noreferrer nofollow" href="https://user-images.githubusercontent.com/541093/195941796-9858e665-dcf4-4838-80a9-954dadea942c.png">source-mobile-nav-open</a></td>
<td><a target="_blank" rel="noopener noreferrer nofollow" href="https://user-images.githubusercontent.com/541093/195941799-d1c920fc-d6ab-46ee-be36-cb87a4552229.png">counter-mobile-nav-open</a></td>
</tr>
</tbody>
</table>

### How to test the changes in this Pull Request:

Visit the 4 download subpages on your local site.

Issue #129, opened 2022-09-28
Download Subpages


javierarce:
Here's the proposed design for the download subpages: Releases, Beta/Nightly, Source Code, Counter that @beafialho, @critterverse, and I have worked on [this file](https://www.figma.com/file/WHhRwVc4swYnPKdyVWj3du/Home-%26-Download-Pages?node-id=2093%3A22538).

I'll share the proposals for the desktop and mobile pages and mention some aspects of the design that could be categorized as 'nice to have'. At the end of the issue I'll add some notes about the navigation of these internal pages:

### Releases 

![image](https://user-images.githubusercontent.com/4933/192808301-1d296cb7-38e8-413b-965c-1542ca298279.png)

![image](https://user-images.githubusercontent.com/4933/192808545-3e6b00d0-7e51-4aea-a27c-cc307a0016f7.png)

![image](https://user-images.githubusercontent.com/4933/192808644-d2ab81bb-9e94-48df-b768-ab5f7ae187ce.png)

### Nice to have

- Instead of listing every release on the same page, older releases could be accessed through a list at the bottom of the page. We could either load each table under the Past releases section or have a separate page per version.
- Each major release is identified with both the version number and its code name.

---

### Beta/Nightly

![image](https://user-images.githubusercontent.com/4933/192808808-aee85f3c-7a93-4a37-af55-bbf0dd858e2d.png)

---

### Counter

![image](https://user-images.githubusercontent.com/4933/192815889-4267f152-b754-404c-a954-8f637bfd43e1.png)

Something we could explore on this page is a smoother transition between numbers. We could also review the font we chose and use a monospaced one so the numbers stay in the same position as the numbers increase.

---

### Source code

![image](https://user-images.githubusercontent.com/4933/192808959-82b380e7-a544-46ba-9833-dc90859ae027.png)

---

### Navigation

![image](https://user-images.githubusercontent.com/4933/192815803-1f4c2ea5-be5a-4c02-9577-dfe6b51aa798.png)

The other challenging aspect of these designs are the subnavigations: on smaller viewports our proposal is for the right-side lists to collapse into dropdowns.

---

[Figma file](https://www.figma.com/file/WHhRwVc4swYnPKdyVWj3du/Home-%26-Download-Pages?node-id=2093%3A22538) with the designs.

ryelle:
[...]

PR #132, opened 2022-10-05
Block: Download Counter

### LINKED ISSUE #130: New Block: Download Counter
Sub-task of #129 

192815889-4267f152-b754-404c-a954-8f637bfd43e1

> Something we could explore on this page is a smoother transition between numbers. We could also review the font we chose and use a monospaced one so the numbers stay in the same position as the numbers increase.

ryelle:
Fixes #130, See #129 — This adds a new block for the Download Counter. It live-updates from a new API endpoint every 5 seconds.

The new API endpoint is `/wporg/v1/core-downloads/[branch]/`, and it uses the same SQL request as the current page. On local sites it uses a workaround to avoid calling a table that doesn't exist, that's described in the code.

There's also a new shortcode, for `stable_branch`, which is a different constant (`WP_CORE_STABLE_BRANCH`) than the existing `latest_release` shortcode. This is used in the heading on the Counter page.

⚠️  This branch is built on #131 (`add/download-subpages`), not `trunk`. The other branch enables the new theme on the download subpages so this can be tested.

### Screenshots

Design-wise, there was a comment about using a monospace font to prevent the numbers jumping around - I went ahead and applied that, but we can still try out different styles. I didn't add any animation, so it just updates to the next number.

https://user-images.githubusercontent.com/541093/194117814-133151c3-399d-4652-9a1a-3f289e9cefb0.mp4

Old version data is still accessible with the `?branch=5.9` query parameter (a current feature of the page).

counter-past

The font scales down on small screens.

counter-small

In the editor, the block doesn't live-update, but it does display the latest count.

editor

This works best on a sandbox, since you can see the real download counts, but it's still possible to test locally.

1. View the Counter page
2. Try passing different branches, ex: `/download/counter/?branch=5.9`
3. Use the block on other pages to test the editor interface — font, color, & spacing should be customizable

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/src/download-counter/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/download-counter/view.js
@@ -19,8 +19,10 @@ const init = () => {
 			}
 
 			setInterval( async () => {
-				const count = await apiFetch( { path: `/wporg/v1/core-downloads/${ branch }` } );
-				element.innerHTML = count;
+				try {
+					const count = await apiFetch( { path: `/wporg/v1/core-downloads/${ branch }` } );
+					element.innerHTML = count;
+				} catch ( error ) {}
 			}, 5000 );
 		} );
 	}
--- a/source/wp-content/themes/wporg-main-2022/src/download-counter/view.js
+++ b/source/wp-content/themes/wporg-main-2022/src/download-counter/view.js
@@ -1,0 +1,29 @@
+/**
+ * WordPress dependencies
+ */
+import apiFetch from '@wordpress/api-fetch';
+
+const init = () => {
+	const containers = document.querySelectorAll( '.wp-block-wporg-download-counter' );
+	if ( containers ) {
+		containers.forEach( ( element ) => {
+			const { branch } = element.dataset;
+
+			// Update the version string if the branch is different, only if the previous element is a heading 1.
+			const heading = element.previousElementSibling;
+			if ( 'H1' === heading.tagName.toUpperCase() ) {
+				const [ version ] = heading.innerHTML.match( /[0-9]+\.[0-9]/ ) || [];
+				if ( version ) {
+					heading.innerHTML = heading.innerHTML.replace( version, branch );
+				}
+			}
+
+			setInterval( async () => {
+				const count = await apiFetch( { path: `/wporg/v1/core-downloads/${ branch }` } );
+				element.innerHTML = count;
+			}, 5000 );
+		} );
+	}
+};
+
+window.addEventListener( 'load', init );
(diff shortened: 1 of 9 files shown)


ryelle:
I had the wrong charcoal color background - it's been updated:

Screen Shot 2022-10-05 at 13 33 38

source/wp-content/themes/wporg-main-2022/src/download-counter/view.js
```diff
@@ -0,0 +1,31 @@
...
+				const [ version ] = heading.innerHTML.match( /[0-9]+\.[0-9]/ ) || [];
+				if ( version ) {
+					heading.innerHTML = heading.innerHTML.replace( version, branch );
+				}
+			}
+
+			setInterval( async () => {
+				try {
+					const count = await apiFetch( { path: `/wporg/v1/core-downloads/${ branch }` } );
+					element.innerHTML = count;
```
StevenDufresne: `innerHTML` seems unnecessary here, probably a better practice to use a safe sink: `element.textContent`.

jasmussen:
Can we use [IBM Plex Mono](https://fonts.google.com/specimen/IBM+Plex+Mono) for the font? Or, @beafialho in case you have thoughts on a better monospace font?

Edit, perhaps 200:

Screenshot 2022-10-06 at 09 51 46

beafialho:
Yes, @jasmussen! On the lack of an Inter monospace version, I think IBM Plex Mono Extralight works perfectly.

ryelle:
@jasmussen @beafialho Should IBM Plex Mono be used anywhere there's a monospace font? Like these URLs on [the Source Code page](https://user-images.githubusercontent.com/541093/194115130-3cf3d3dc-feeb-4cce-a394-212308c1d793.png)? Or is this just for this one block?

jasmussen:
It's a beautiful font, I'd love to see it used for any and all monospace cases. But I'll defer to Bea!

beafialho:
> @jasmussen @beafialho Should IBM Plex Mono be used anywhere there's a monospace font?

I agree, this can be used everywhere. In the case of these URL's though, the weight used would be `Regular 400`.

ryelle:
Okay, I've created https://github.com/WordPress/wporg-mu-plugins/issues/290 & https://github.com/WordPress/wporg-parent-2021/issues/47 for adding the new font. I'll make sure that gets applied before the page is launched, but I don't think those need to block merging this block. Is there any other design feedback?

ryelle:
Updated to use IBM Plex Mono:

Screen Shot 2022-10-06 at 17 01 18

StevenDufresne (commented):
I've tested the code, works as expected.

The JS looks good, the PHP also looks fine although you could loop someone else for a more detailed review.

I'm not sure if there are more design iterations, so I'll leave approval to others.

jasmussen:
Looks good!

beafialho:
@ryelle can we use the Extra Light weight for big text such as counters, etc? Regular should be used for small text.

Captura de ecrã 2022-10-07, às 09 37 51

PR #134, opened 2022-10-12
Block: Release Tables

ryelle:
See #129 — This adds a new block for all the release tables on https://wordpress.org/download/releases/.

It displays the latest release, then the last two version branches. Below that, it has a list of past versions, betas, and the MU releases. Clicking one of those toggles the table to be visible. This is using a tab-panel pattern for accessibility. It also updates and reads from the URL hash, so if you link to `https://wordpress.org/download/releases/#betas`, the beta table will be displayed.

I haven't changed anything about the source data, so we don't have the release code name. We can iterate on adding that later.

⚠️  This branch is built on #131 (`add/download-subpages`), not `trunk`. The other branch enables the new theme on the download subpages so this can be tested.

### Screenshots

| Desktop | Mobile | Editor
|--------|-------|-------|
| releases-1200 | releases-480 | releases-editor |

releases-old-open

1. Make sure your `wporg-parent-2021` is up to date (the default table styling has been updated)
2. If testing locally, you'll need to mock the release data — drop [this file](https://gist.github.com/ryelle/5a36ea8090e1bf095a1c5b5496561786) into your `mu-plugins` folder.
3. View the releases page
4. There should be tables matching the screenshots
5. The links under "Past releases" should work to toggle between older version tables

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -9,6 +9,7 @@ require_once( __DIR__ . '/inc/capabilities.php' );
 // Block files
 require_once( __DIR__ . '/src/download-counter/index.php' );
 require_once( __DIR__ . '/src/random-heading/index.php' );
+require_once( __DIR__ . '/src/release-tables/index.php' );
 
 /**
  * Actions and filters.
--- a/source/wp-content/themes/wporg-main-2022/patterns/releases.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/releases.php
@@ -30,6 +30,8 @@
 
 <!-- wp:paragraph -->
 <p><?php _e( 'Curious about which jazzers we highlighted for each release? <a href="https://wordpress.org/about/history/">It’s on the History page</a>.', 'wporg' ); ?></p>
-<!-- /wp:paragraph --></div>
+<!-- /wp:paragraph -->
+
+<!-- wp:wporg/release-tables /--></div>
 <!-- /wp:column --></div>
 <!-- /wp:columns -->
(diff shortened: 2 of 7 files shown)


javierarce:
This looks great! I'll try to test the branch locally, but in the meantime I just saw a couple of minor things:

- Could we say Releases instead of Release Archive? Besides matching the text with the links to this page, a shorter string will help with the translation to other languages.
-  Instead of separating the hashes with a bar (md5 | sha1) could we use a dot (md5 · sha1)?

> It also updates and reads from the URL hash, so if you link to https://wordpress.org/download/releases/#betas, the beta table will be displayed.

❤️

ryelle:
> Could we say Releases instead of Release Archive? 

Yep, Release Archive was the current page title — I believe I already made that change in the page sync PR (#131), this one's just for the table itself.

> Instead of separating the hashes with a bar (md5 | sha1) could we use a dot (md5 · sha1)?

Done:
Screen Shot 2022-10-13 at 10 14 49 AM

adamwoodnz (approved):
This is really nice, learned a lot reading through.

Keyboard nav works well on tablist. Voiceover leaves me on the selected tab rather than moving focus to the table, not sure what's expected here.

I was a little surprised that we'd have to roll our own tabs component. I wonder if this could be made reusable... but maybe that can happen the next time we need tabs.

LGTM anyway 🌟

ryelle:
> Voiceover leaves me on the selected tab rather than moving focus to the table, not sure what's expected here.

I had that too, I figured I would ask the #accessibility team for feedback once the pages are live (so they're easier to test).

> I wonder if this could be made reusable... but maybe that can happen the next time we need tabs.

Yep, that works for me 👍🏻

PR #135, opened 2022-10-13
Bypass the sandbox when running build:patterns

tellyworth:
This uses `wp_remote_get()` and the `CURLOPT_CONNECT_TO` curl option to ensure `yarn build:patterns` will always connect to production servers, even if `wordpress.org` is sandboxed on your local machine.

Without this, Docker will fail to connect when you're sandboxed, because it respects /etc/hosts but doesn't use the proxy, so the request is denied.

See https://github.com/WordPress/wporg-main-2022/pull/118#discussion_r968067740

1. Sandbox wordpress.org in `/etc/hosts`.
2. `yarn build:patterns` should work fine with this PR applied.
3. Try without the PR, you'll get something like a `cURL error 28: Failed to connect` timeout.

### DIFF
--- a/env/export-content/includes/utils.php
+++ b/env/export-content/includes/utils.php
@@ -6,10 +6,29 @@ namespace WordPress_org\Main_2022\ExportToPatterns;
 require __DIR__ . '/parser.php';
 
 /**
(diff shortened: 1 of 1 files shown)


ryelle (approved):
👍🏻

PR #138, opened 2022-10-20
Add translator strings for preloading required fonts

### LINKED ISSUE #136: Preload required font subsets for rosetta sites once we support them
See https://github.com/WordPress/wporg-main-2022/pull/103/commits/93a44fa9155caad28bc76e46a565cedd6c24bfc6#diff-b4f8b812eb373b4d5dd73ee10679133902227145fb8666a0d87cb282ed5bb51eR67

Currently, we specifically specify the preloaded font subsets as `Latin` for [wporg-main homepage heading](https://github.com/WordPress/wporg-main-2022/pull/103/commits/7696aa3b573a4e0901ac0898f5b4da45e0f36fe0). In the future, we also want to preload the required fonts for the rosetta sites once they are supported.

outdoor2kode:
Fixes #136 

Translators now have the option to add necessary fonts for preloading to rosetta sites.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -66,14 +66,19 @@ function enqueue_assets() {
 
 	// Preload the heading font(s).
 	if ( is_callable( 'global_fonts_preload' ) ) {
-		// TODO: Load required fonts for rosetta sites once they are supported.
-		if ( ! is_rosetta_site() ) {
-			// All headings.
-			global_fonts_preload( 'EB Garamond Latin' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+		$heading_subsets = explode( ',', _x( 'latin', 'Heading font subsets, comma separated', 'wporg' ) );
+		// All headings.
+		foreach ( $heading_subsets as $heading_subset ) {
+			global_fonts_preload( 'EB Garamond', $heading_subset );
+		}
 
-			if ( is_front_page() ) {
-				// The heading on the front-page has some italic.
-				global_fonts_preload( 'EB Garamond Latin italic' );
+		if ( is_front_page() ) {
+			/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+			$front_page_heading_subsets = explode( ',', _x( 'latin', 'Front page heading font subsets, comma separated', 'wporg' ) );
+			// The heading on the front-page has some italic.
+			foreach ( $front_page_heading_subsets as $front_page_heading_subset ) {
+				global_fonts_preload( 'EB Garamond italic', $front_page_heading_subset );
 			}
 		}
 	}
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -67,19 +67,13 @@ function enqueue_assets() {
 	// Preload the heading font(s).
 	if ( is_callable( 'global_fonts_preload' ) ) {
 		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
-		$heading_subsets = explode( ',', _x( 'latin', 'Heading font subsets, comma separated', 'wporg' ) );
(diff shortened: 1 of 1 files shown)


source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -65,14 +65,19 @@ function enqueue_assets() {
...
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+		$heading_subsets = explode( ',', _x( 'latin', 'Heading font subsets, comma separated', 'wporg' ) );
+		// All headings.
+		foreach ( $heading_subsets as $heading_subset ) {
+			global_fonts_preload( 'EB Garamond', $heading_subset );
+		}
+
+		if ( is_front_page() ) {
+			/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+			$front_page_heading_subsets = explode( ',', _x( 'latin', 'Front page heading font subsets, comma separated', 'wporg' ) );
```
ryelle: I think we could probably assume the subsets needed for the italic text would be the same as the general subsets, right?

source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -65,14 +65,19 @@ function enqueue_assets() {
...
-			global_fonts_preload( 'EB Garamond Latin' );
-
-			if ( is_front_page() ) {
-				// The heading on the front-page has some italic.
-				global_fonts_preload( 'EB Garamond Latin italic' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+		$heading_subsets = explode( ',', _x( 'latin', 'Heading font subsets, comma separated', 'wporg' ) );
+		// All headings.
+		foreach ( $heading_subsets as $heading_subset ) {
+			global_fonts_preload( 'EB Garamond', $heading_subset );
```
ryelle: Do you think we should also call for preloading Inter here, just in case it's ever removed from the global header? Or is that unlikely enough to not bother?

source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -65,14 +65,19 @@ function enqueue_assets() {
...
-		// TODO: Load required fonts for rosetta sites once they are supported.
-		if ( ! is_rosetta_site() ) {
-			// All headings.
-			global_fonts_preload( 'EB Garamond Latin' );
-
-			if ( is_front_page() ) {
-				// The heading on the front-page has some italic.
-				global_fonts_preload( 'EB Garamond Latin italic' );
+		/* translators: Subsets can be any of cyrillic, cyrillic-ext, greek, greek-ext, vietnamese, latin, latin-ext. */
+		$heading_subsets = explode( ',', _x( 'latin', 'Heading font subsets, comma separated', 'wporg' ) );
```
ryelle: Is `explode` necessary here? `global_fonts_preload` expects the `$subsets` parameter to be a comma-separated string, not an array.

outdoor2kode: It was added because of the original function design, `global_fonts_preload` can't handle cases like `'fonts' => [ 'Inter' ], 'subsets' => [ 'Cyrillic, Cyrillic-ext' ]` (`Cyrillic-ext` can't match anything. See https://github.com/WordPress/wporg-mu-plugins/pull/298#discussion_r1013407403). For such cases, it needs to call `global_fonts_preload( 'Inter, Inter', 'Cyrillic, Cyrillic-ext' );` So I `explode` and `foreach` it here.

With the commit [Update preload_font logic. See README for details.](https://github.com/WordPress/wporg-mu-plugins/pull/297/commits/7851a09f6dde6db8684893ac1cff25aa35c30330), `explode` is removed.

source/wp-content/themes/wporg-main-2022/functions.php
outdoor2kode: Yeah, good point. Thanks!

source/wp-content/themes/wporg-main-2022/functions.php
outdoor2kode: Hmm, I think it's ok not to preload `Inter` here as compared to the purpose of preloading `EB Garamond` here - avoid the text shaking when loading the page), I don't see enough benefits to do so for `Inter`. Do you think it makes sense?

ryelle (approved):

PR #151, opened 2022-11-28
Update the layout, position and font size of homepage callout.

StevenDufresne:
This Pull Request is driven by a recent design change made in the wporg-showcase redesign where the content is now left aligned and the font size is `20px`. ([Showcase Figma](https://www.figma.com/file/nK6LcMOZMtRAQdIky3PNIA/Showcase-Pages?node-id=1274%3A21566&t=TtkEaadC1gRm7UYc-0)).

Related https://github.com/WordPress/wporg-showcase-2022/pull/60.

### Screenshots

| Before | After | 
| -------| -------| 
|  wordpress org_ | localhost_8888_ |

There are no visual changes for mobile & tablet views.

1. Pull changes, load homepage.


source/wp-content/themes/wporg-main-2022/patterns/home.php
```diff
@@ -280,15 +280,15 @@
...
 <div class="wp-block-group alignwide"><!-- wp:heading {"align":"wide","style":{"typography":{"fontStyle":"normal","fontWeight":"300","letterSpacing":"-2px"}},"className":"is-style-with-arrow","fontSize":"heading-cta","fontFamily":"inter"} -->
 <h2 class="alignwide is-style-with-arrow has-inter-font-family has-heading-cta-font-size" style="font-style:normal;font-weight:300;letter-spacing:-2px"><?php _e( '<a href="https://wordpress.org/download/">Get started</a>', 'wporg' ); ?></h2>
 <!-- /wp:heading -->
 
 <!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":"0px"}}} -->
 <div class="wp-block-columns alignwide"><!-- wp:column {"width":"50%"} -->
-<div class="wp-block-column" style="flex-basis:50%"><!-- wp:paragraph {"className":"is-style-short-text"} -->
+<div class="wp-block-column" style="flex-basis:50%"><!-- wp:paragraph {"className":"is-style-short-text","fontSize":"large"} -->
```
adamwoodnz, line 291: I'm not seeing this large font size applied for some reason. It's displaying as 16px, and there is no corresponding classname on the node:

Screen Shot 2022-12-01 at 12 12 47 PM

Same for you?

Adding `has-large-font-size` classname in the browser does bump it to 20px so I assume that's what should be applied.

The constrained layout and left alignment is working so I'm pretty sure the pattern is being loaded correctly.

adamwoodnz (commented):
Alignment is working for me but not the font size change...

source/wp-content/themes/wporg-main-2022/patterns/home.php
StevenDufresne: I see what's going on.  I added the property to the template, and opened the block editor and it works in the block editor without the CSS class applied. I must have stopped short of checking the front. 🙇

outdoor2kode (approved):
Tested and Looks good to me 👍

ryelle (approved):
Technically this looks fine. Should this be updated on all sites that use this pattern (including in the parent pattern)? Since it's also used on Enterprise, Documentation, etc.

StevenDufresne:
Showcase was reverted. I think I'll close this. We can reopen if we want to make that update.

PR #152, opened 2022-12-01
Ensure Designers can only edit pages that already have a block template

tellyworth:
Editing a page without a block template means changes would immediately go live. This prevents it entirely, which means the workflow requires a dev to create the page template before a designer can start editing.

1. Log in as a Designer user
2. Go to wp-admin / Pages / All Pages
3. Pages that have templates (like Download) should have an Edit button; others (like Enterprise currently) should not.


ryelle:
Looking at the code, this will work to allow designer-users to only edit pages, but they won't be able to build up new pages (like Enterprise, or About). Ideally the designer-user should still be able to edit those pages, but it should not change the live page's `_wp_page_template`. Even if a developer creates an empty page template, the meta value will still be wiped out when anyone saves the existing page, since for example, the `page-enterprise.php` template is not available in the new theme.

PR #155, opened 2022-12-08
Add sotw banner to front page pattern

tellyworth:
I've added this above the survey banner; is that correct?

### Screenshots

Screen Shot 2022-12-08 at 6 31 06 pm

Screen Shot 2022-12-08 at 6 31 22 pm

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/home.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/home.php
@@ -6,6 +6,28 @@
  */
 
 ?>
+<!-- wp:group {"align":"full","style":{"color":{"background":"#020c3a"}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull has-background" style="background-color:#020c3a"><!-- wp:cover {"url":"https://wordpress.org/files/2022/12/sotw-drawer-background.png","id":14034,"dimRatio":0,"minHeight":350,"align":"wide"} -->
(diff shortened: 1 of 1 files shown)


thetinyl:
Thanks @tellyworth! To clarify and confirm: the State of the Word banner will be replacing the survey banner entirely. The survey banner will be swapped back in a week or so after the State of the Word event.

source/wp-content/themes/wporg-main-2022/patterns/home.php
```diff
@@ -6,6 +6,28 @@
  */
 
 ?>
+<!-- wp:group {"align":"full","style":{"color":{"background":"#020c3a"}},"layout":{"type":"constrained"}} -->
```
iandunn, line 10: It could be convenient to add an `if ( time() < strtotime( '2022-12-22' ) )` condition around this (and the reverse condition for removing the SotW content).

That way we they'll be added/removed at the right times automatically, rather than us having to manually make another pr/deploy.

tellyworth: Good suggestion @iandunn, thanks. I'm going to deploy a change with a php conditional manually included. It's a departure from the usual process, so I'm curious to hear what @ryelle thinks when she's back. It can be overwritten by a different change when necessary.

Long term, it would be useful to have conditional blocks so we can do this kind of thing without needing php code.

ryelle: I think we can do that for specific cases (like this, when it's very important that something be shown/hidden at a certain time), but generally this file shouldn't be edited manually — it's generated from the live page content via `yarn build:patterns`. Running that again if we needed to update the copy, for example, would blow away this conditional.

> Long term, it would be useful to have conditional blocks so we can do this kind of thing without needing php code.

Yeah, an editor-first solution would be better.

PR #156, opened 2022-12-09
Rename to `wp-cli.yml` to allow override

iandunn:
The `wp-cli.local.yml` file was introduced in #35.

I think `wp-cli.yml` is intended to be analogous to `phpunit.xml.dist`, which establishes defaults for a repository. Then, `wp-cli.local.yml` is analogous  to `phpunit.xml`, which can override the `.dist` file.

So, the `local` file shouldn't be committed to source, because that would prevent local envs from overriding it when they need to (like if they're not using the Docker env).

This change shouldn't break compat, since the new `wp-cli.yml` will still take effect. The only change is that someone can now setup a `wp-cli.local.yml` file if they need to.


ryelle (approved):
The change makes sense to me 👍🏻

PR #157, opened 2022-12-14
Add template, pattern and styles for Enterprise page

### LINKED ISSUE #150: Add new Enterprise content and design
This design is still in progress, not yet final.

enterprise-i4 7

(updated with newer design)

adamwoodnz:
Closes #150 

Adds a template and pattern for the Enterprise page. Some custom styles have had to be added, mostly for the logo grid which changes layout significantly on mobile.

Remaining tasks:
- [x] Prod quality images
- [x] Escaping
- [x] Arrow styling for last section
- [x] Heading hierarchy
- [x] A11y
- [x] Lighthouse Perf and a11y tests

Props @bengreeley

### Screenshots

| Desktop | Tablet | Mobile |
|--------|-------|-------|
| localhost_8888_enterprise_ | localhost_8888_enterprise_(iPad) | localhost_8888_enterprise_(Samsung Galaxy S20 Ultra) |

1. Add an 'Enterprise' page with slug `enterprise` (an existing page may need to be deprecated in your site)
2. Rebuild the styles `yarn workspace wporg-main-2022-theme build`
3. Test the responsive layout

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
@@ -20,8 +20,8 @@
 <!-- /wp:heading --></div>
 <!-- /wp:column -->
 
-<!-- wp:column {"verticalAlignment":"center","width":"45%","style":{"spacing":{"padding":{"bottom":"var:preset|spacing|40","top":"var:preset|spacing|60","right":"var:preset|spacing|60","left":"var:preset|spacing|60"}}}} -->
-<div class="wp-block-column is-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--60);flex-basis:45%"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"center"}} -->
+<!-- wp:column {"verticalAlignment":"center","width":"45%","style":{"spacing":{"padding":{"bottom":"var:preset|spacing|40","top":"var:preset|spacing|60","right":"var:preset|spacing|edge-space","left":"var:preset|spacing|edge-space"}}}} -->
+<div class="wp-block-column is-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--edge-space);flex-basis:45%"><!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"center"}} -->
 <div class="wp-block-group"><!-- wp:paragraph {"align":"center","placeholder":"\u003c?php _e( 'Content…', 'wporg' ); ?\u003e","style":{"typography":{"lineHeight":1.7}},"textColor":"white","className":"is-style-default","fontSize":"small"} -->
 <p class="has-text-align-center is-style-default has-white-color has-text-color has-small-font-size" style="line-height:1.7"><?php _e( 'WordPress content management system has grown to become a dominant market leader. Discover how today&#039;s biggest brands use WordPress, the features that make it so appealing and how you can make WordPress work for your enterprise.', 'wporg' ); ?></p>
 <!-- /wp:paragraph --></div>
@@ -78,16 +78,15 @@
 <!-- /wp:group --></div>
 <!-- /wp:group -->
 
-<!-- wp:group {"align":"full","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull"><!-- wp:columns {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|60","left":"0"}}}} -->
-<div class="wp-block-columns alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-left:0"><!-- wp:column {"verticalAlignment":"bottom","width":"33%","className":"wporg-hidden-mobile"} -->
-<div class="wp-block-column is-vertically-aligned-bottom wporg-hidden-mobile" style="flex-basis:33%"><!-- wp:image {"width":402,"height":588,"sizeSlug":"large"} -->
-<figure class="wp-block-image size-large is-resized"><img src="https://wordpress.org/files/2022/12/Enterprise-Use-Cases.png" alt="<?php _e( 'Enterprise Use Cases Image', 'wporg' ); ?>" width="402" height="588" /></figure>
-<!-- /wp:image --></div>
+<!-- wp:cover {"url":"https://wordpress.org/files/2022/12/Enterprise-Use-Cases.png","id":14920,"dimRatio":0,"focalPoint":{"x":0,"y":1},"isDark":false,"align":"full","className":"wporg-enterprise-use-cases","style":{"spacing":{"padding":{"top":"0","right":"0","bottom":"0","left":"0"}}}} -->
+<div class="wp-block-cover alignfull is-light wporg-enterprise-use-cases" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><span aria-hidden="true" class="wp-block-cover__background has-background-dim-0 has-background-dim"></span><img class="wp-block-cover__image-background wp-image-14920" alt="" src="https://wordpress.org/files/2022/12/Enterprise-Use-Cases.png" style="object-position:0% 100%" data-object-fit="cover" data-object-position="0% 100%" /><div class="wp-block-cover__inner-container"><!-- wp:group {"align":"full","layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull"><!-- wp:columns {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|edge-space","left":"var:preset|spacing|edge-space"}}}} -->
+<div class="wp-block-columns alignwide" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--edge-space);padding-left:var(--wp--preset--spacing--edge-space)"><!-- wp:column {"verticalAlignment":"bottom","width":"33%"} -->
+<div class="wp-block-column is-vertically-aligned-bottom" style="flex-basis:33%"></div>
 <!-- /wp:column -->
 
-<!-- wp:column {"width":"67%","style":{"spacing":{"padding":{"left":"var:preset|spacing|60","right":"0","top":"0","bottom":"0"}}}} -->
-<div class="wp-block-column" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:var(--wp--preset--spacing--60);flex-basis:67%"><!-- wp:heading {"style":{"spacing":{"margin":{"top":"0","right":"0","bottom":"0","left":"0"}}},"className":"wp-block-heading","fontSize":"heading-3"} -->
+<!-- wp:column {"width":"67%"} -->
+<div class="wp-block-column" style="flex-basis:67%"><!-- wp:heading {"style":{"spacing":{"margin":{"top":"0","right":"0","bottom":"0","left":"0"}}},"className":"wp-block-heading","fontSize":"heading-3"} -->
 <h2 class="wp-block-heading has-heading-3-font-size" style="margin-top:0;margin-right:0;margin-bottom:0;margin-left:0"><?php _e( 'Enterprise use cases', 'wporg' ); ?></h2>
 <!-- /wp:heading -->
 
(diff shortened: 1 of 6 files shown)


source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
```diff
@@ -0,0 +1,266 @@
...
+ * Inserter: no
+ */
+
+?>
+
+<!-- wp:group {"align":"full","backgroundColor":"charcoal-2","className":"wporg-enterprise-header","layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull wporg-enterprise-header has-charcoal-2-background-color has-background"><!-- wp:columns {"verticalAlignment":"center","align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|90"}}}} -->
+<div class="wp-block-columns alignfull are-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--90)"><!-- wp:column {"verticalAlignment":"center","width":"55%","style":{"spacing":{"padding":{"left":"var:preset|spacing|10","right":"var:preset|spacing|20"}}}} -->
+<div class="wp-block-column is-vertically-aligned-center" style="padding-right:var(--wp--preset--spacing--20);padding-left:var(--wp--preset--spacing--10);flex-basis:55%"><!-- wp:image {"sizeSlug":"large"} -->
+<figure class="wp-block-image size-large"><img src="https://wordpress.org/files/2022/12/WordPress-Enterprise.png" alt="<?php _e( 'WordPress Enterprise', 'wporg' ); ?>" width="100%" height="100%"/></figure>
```
StevenDufresne: This `width` & `height` on the element is causing issues in the block editor I beleive.

 
Screen Shot 2022-12-15 at 2 21 54 PM

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
```diff
@@ -0,0 +1,266 @@
...
+<!-- /wp:group --></div>
+<!-- /wp:group --></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"align":"full","layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull"><!-- wp:columns {"verticalAlignment":"top","align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|60","left":"0"}}}} -->
+<div class="wp-block-columns alignfull are-vertically-aligned-top" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-left:0"><!-- wp:column {"width":"33%","className":"wporg-hidden-mobile"} -->
+<div class="wp-block-column wporg-hidden-mobile" style="flex-basis:33%">
+<!-- wp:image {"width":402,"height":588,"sizeSlug":"large"} -->
+<figure class="wp-block-image size-large is-resized"><img src="https://wordpress.org/files/2022/12/Enterprise-Use-Cases.png" alt="<?php _e( 'Enterprise Use Cases Image', 'wporg' ); ?>" width="402" height="588"/></figure>
```
StevenDufresne: This image is naturally cut off on the bottom. When the screen shrinks, before it is hidden, the cut off is visible.

`~1000px`
Screen Shot 2022-12-15 at 2 24 36 PM

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
```diff
@@ -0,0 +1,266 @@
...
+<!-- /wp:column -->
+
+<!-- wp:column {"className":"wp-block-navigation"} -->
+<div class="wp-block-column wp-block-navigation"><!-- wp:heading {"level":3,"className":"wp-block-heading","fontSize":"heading-5"} -->
+<h3 class="wp-block-heading has-heading-5-font-size"><?php _e( 'White paper references', 'wporg' ); ?></h3>
+<!-- /wp:heading -->
+
+<!-- wp:list {"fontSize":"small"} -->
+<ul class="has-small-font-size"><!-- wp:list-item -->
+<li><a href="https://wordpress.org/files/2020/11/IDC-Report-Choosing-a-CMS-to-Meet-Todays-Digital-Experience-Challenges.pdf" data-type="URL" data-id="https://wordpress.org/files/2020/11/IDC-Report-Choosing-a-CMS-to-Meet-Todays-Digital-Experience-Challenges.pdf"><?php _e( 'IDC Report (PDF)', 'wporg' ); ?>↗</a></li>
```
StevenDufresne: We need to be consistent with our "external" link icon. It appears like we have used that north-east arrow for external links here, but not on `https://wordpress.org/gutenberg/` or in the state of the word address banner currently on the homepage.

As per mat's request to have all external links on .org denoted, I implemented an external link class in https://github.com/WordPress/wporg-parent-2021/pull/53, assuming that's how we wanted to denote them (And how Gutenberg does). But if that's not the case, and we want to use `↗`,  let's figure that out so I can update: https://github.com/WordPress/wporg-parent-2021/pull/53. We currently are only using it on one link in the showcase.

@jasmussen @javierarce @beafialho

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
```diff
@@ -0,0 +1,266 @@
...
+<div class="wp-block-column" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-width:1px;border-right-color:var(--wp--preset--color--light-grey-1);border-right-width:1px;border-bottom-style:none;border-bottom-width:0px;border-left-style:none;border-left-width:0px;padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)">
+<!-- wp:image {"id":13887,"width":58,"height":60,"sizeSlug":"full","linkDestination":"none"} -->
+<figure class="wp-block-image size-full is-resized"><img src="https://wordpress.org/files/2022/11/icon-extensibility.png" alt="<?php _e( 'Extensibility icon', 'wporg' ); ?>" class="wp-image-13887" width="58" height="60"/></figure>
+<!-- /wp:image -->
+
+<!-- wp:heading {"level":4,"className":"wp-block-heading"} -->
+<h4 class="wp-block-heading"><?php _e( 'Extensibility', 'wporg' ); ?></h4>
```
StevenDufresne: Should these be level 2/3?

Screen Shot 2022-12-15 at 2 40 29 PM

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
```diff
@@ -0,0 +1,266 @@
...
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)"><!-- wp:heading {"align":"full","style":{"typography":{"fontStyle":"normal","fontWeight":"300","letterSpacing":"-2px"}},"className":"wp-block-heading is-style-with-arrow","fontSize":"heading-cta","fontFamily":"inter"} -->
+<h2 class="wp-block-heading alignfull is-style-with-arrow has-inter-font-family has-heading-cta-font-size" style="font-style:normal;font-weight:300;letter-spacing:-2px"><a href="https://wordpress.org/download/"><?php _e( 'Try WordPress', 'wporg' ); ?></a></h2>
```
StevenDufresne: This components stretches pretty wide. 

Not sure if that's what we want but in the showcase, this component stops at `1160px`

**Showcase**
Screen Shot 2022-12-15 at 2 47 28 PM
**This PR**
Screen Shot 2022-12-15 at 2 46 24 PM

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
```diff
@@ -0,0 +1,266 @@
...
+<div class="wp-block-group alignfull wporg-enterprise-header has-charcoal-2-background-color has-background"><!-- wp:columns {"verticalAlignment":"center","align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|90"}}}} -->
+<div class="wp-block-columns alignfull are-vertically-aligned-center" style="padding-top:var(--wp--preset--spacing--90)"><!-- wp:column {"verticalAlignment":"center","width":"55%","style":{"spacing":{"padding":{"left":"var:preset|spacing|10","right":"var:preset|spacing|20"}}}} -->
+<div class="wp-block-column is-vertically-aligned-center" style="padding-right:var(--wp--preset--spacing--20);padding-left:var(--wp--preset--spacing--10);flex-basis:55%"><!-- wp:image {"sizeSlug":"large"} -->
+<figure class="wp-block-image size-large"><img src="https://wordpress.org/files/2022/12/WordPress-Enterprise.png" alt="<?php _e( 'WordPress Enterprise', 'wporg' ); ?>" width="100%" height="100%"/></figure>
+<!-- /wp:image -->
+
+<!-- wp:heading {"level":1,"className":"screen-reader-text"} -->
+<h1 class="wp-block-heading screen-reader-text"><?php _e( 'WordPress Enterprise', 'wporg' ); ?></h1>
```
StevenDufresne, line 19: I just want to mention that this uses the same approach that showcase currently does which was flagged by a11y.


https://github.com/WordPress/wporg-showcase-2022/issues/89

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: Ah I added that last minute to satisfy Lighthouse, will remove, thanks

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: Happy to go with the preferred pattern here, seems like the class is?

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: Yeah it wasn't clear to me from the design; it appears like everything should stretch full width, but maybe it should be consistent with other sites rather than the sections above.

Screen Shot 2022-12-16 at 9 32 20 AM

@jasmussen @javierarce @beafialho ?

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: I decided to be consistent until we have an approach using a webfont

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: This could be approached in 2 ways; the simple way is to position it at the bottom, always cut off and let it shrink down into the corner. The more complex way is to use the full image, positioned centrally with a slight translation lower and overflow hidden, and as it shrinks more of the image will come into view.

For now I've updated to the simple way.

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: Removed

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: Fixed

Screen Shot 2022-12-16 at 12 23 07 PM

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: It would be really helpful actually if the designs showed a wider width so that we can tell if the layout is constrained or not. The current page design width is 1320px with the content at 1166px, which suggests to me that maybe all, or at least some of, these sections should be constrained, but I can't tell.

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
StevenDufresne: I don't know. I'm still trying to lock down the experience:

https://github.com/WordPress/wporg-showcase-2022/pull/110#issuecomment-1353874722

The design team has the final say of course :)

javierarce: We are reviewing some decisions about the use of the arrow icons to remove the discrepancies. 

We have updated the designs [here](https://www.figma.com/file/QK1zsWPQMTEYuar0SLepsI/Enterprise-Page?node-id=1%3A2193) to just use the classic icon for external links (and nothing on the internal ones).

image

cc: @panchovm

fcoveram: Based on what @javierarce shared above, we think the best approach is to use the [external link icon](https://github.com/WordPress/wporg-parent-2021/issues/64#issuecomment-1346174965) replacing the `↗︎` glyph in all external links.

On the other hand, it would be good to set all subdomain links as internal links and prevent showing the icon repeatedly in sections that don't need it, e.g. wp.org footer.

CleanShot 2022-12-16 at 12 50 41@2x

cc: @beafialho @StevenDufresne @bengreeley

adamwoodnz: Replaced the characters with the SVG icon via the `external-link` class, note that the icon is `--wp--preset--color--charcoal-4` so we may need to do more work there

Screen Shot 2022-12-19 at 10 54 26 AM

adamwoodnz: I see the color change is mentioned here https://github.com/WordPress/wporg-showcase-2022/pull/110#issuecomment-1354572161

adamwoodnz:
Updated screenshots below.

@javierarce @panchovm please let me know how you'd like the layout to behave at really wide sizes.

As you can see, currently at 1600px we're letting everything continue to stretch, but we could enforce the wide constrained width on some sections. The 'Use Cases' section will require custom CSS to position the left image correctly, as I haven't managed to find a way to use a constrained layout with the image remaining positioned to the bottom left in the editor using Cover or Media + Text.

| 1600px | 1320px (as designed) | Tablet | Mobile |
|--------|-------|-------|--------|
| localhost_8888_enterprise-2_| localhost_8888_enterprise-2_ (1) | localhost_8888_enterprise-2_(iPad) | localhost_8888_enterprise-2_(Samsung Galaxy S20 Ultra) |

fcoveram:
I would follow the same approach as in the homepage. Setting a maximum width in the main grid, where text content is placed, and scaling up the image cropped by the section.

Fixing the image in the bottom left would add white space between the self and right side, breaking the three columns grid generated by being next to the Features section. Sharing a quick draft I made to see how it might work out.

CleanShot 2022-12-20 at 09 51 09@2x

|Fixed image|Scaled up image|
|:---|:---|
|CleanShot 2022-12-20 at 10 00 57@2x|CleanShot 2022-12-20 at 09 51 09@2x|

---

On the other hand, I am noticing from the Mobile screenshot that "Try WordPress" section has a different letter-spacing value. I checked the same section in homepage by replacing the label, and the outcome looks less tight. I think we should follow that same style. Adding some screenshots

|Section in Mobile screenshot|Section in Homepage (`320px` viewport)|
|:---|:---|
CleanShot 2022-12-20 at 09 55 23@2x|CleanShot 2022-12-20 at 09 56 32@2x|

adamwoodnz:
> "Try WordPress" section has a different letter-spacing value

I had to reduce the letter spacing to `-6px` to make the text fit on one line on wide screens. At `-2px` it looks like this:

Screen Shot 2023-01-05 at 12 04 00 PM

We could of course use different values on mobile and desktop, or reduce the font size. Let me know what you'd prefer @panchovm

adamwoodnz:
Updated screenshots below with constrained layout and 'Try WordPress' font size reduced to fit (99px, down from 120px), rather than using letter spacing.

@javierarce @panchovm please let me know what you think.

| 1600px | 1320px (as designed) | Tablet | Mobile |
|--------|-------|-------|--------|
| localhost_8888_enterprise-2_ (3) | localhost_8888_enterprise-2_ (2) | localhost_8888_enterprise-2_(iPad) (1) | localhost_8888_enterprise-2_(Samsung Galaxy S20 Ultra) (1) |

StevenDufresne:
This looks good. 👍 

I do find that the "WordPress Enterprise" graphic gets blurry on my large desktop.

adamwoodnz:
> I do find that the "WordPress Enterprise" graphic gets blurry on my large desktop.

Yeah agreed, nice catch, I'll bump it up a bit more.

source/wp-content/themes/wporg-main-2022/patterns/enterprise.php
adamwoodnz: Color has been updated

source/wp-content/themes/wporg-main-2022/src/style/style.scss
```diff
@@ -60,6 +60,27 @@ html[dir="rtl"] :where([style*="border-right-color"]) {
 	background-color: var(--wp--preset--color--charcoal-1);
 }
 
+/**
+ * Enterprise page
+ */
+
+.wporg-enterprise {
+	&-header .wp-block-columns {
```
ryelle: We have a style rule about this, but I try to avoid using the `&` in selector names. I think it makes it harder to look at a single ruleset and know where it applies. Do you think it's useful here?

source/wp-content/themes/wporg-main-2022/src/style/style.scss
```diff
@@ -60,6 +60,27 @@ html[dir="rtl"] :where([style*="border-right-color"]) {
...
+		gap: 0;
```
ryelle: Can you use the block spacing setting for this? If you split it, you can use "vertical" to apply on mobile.

Screenshot 2023-01-05 at 5 53 40 PM

ryelle (commented):
Two pretty minor comments from a quick review, otherwise page looks good 👍🏻

source/wp-content/themes/wporg-main-2022/src/style/style.scss
adamwoodnz: Hmm I can see it might be problematic with a large number of nested selectors, where you've scrolled down far enough to lose sight of the prefix. In a small block like this, I like that it's DRY and it acts as a wrapper for the styles for this page. I guess we have comments for that though. Happy to refactor. Should it be failing lint? Could you please point me to the rule?

source/wp-content/themes/wporg-main-2022/src/style/style.scss
adamwoodnz: I would love to but I don't have the option to split, locally or in prod

Screen Shot 2023-01-06 at 3 08 16 PM

adamwoodnz: Ignore, I was looking at a row block, not the columns. Updated.

source/wp-content/themes/wporg-main-2022/src/style/style.scss
ryelle: Sorry, I mis-typed 🤦🏻  — we _don't_ have a style rule about this.

You're right that here it's a small enough section to not be a real problem, but I prefer to avoid this pattern always to remove the subjective "is this short enough" decision. I would appreciate a refactor.

I started an issue to discussing adding a rule to our project lint config https://github.com/WordPress/wporg-repo-tools/issues/25

source/wp-content/themes/wporg-main-2022/src/style/style.scss
ryelle: I did the same thing when trying to take that screenshot 😁

source/wp-content/themes/wporg-main-2022/src/style/style.scss
adamwoodnz: Code refactored and I commented on the issue.

fcoveram (changes requested):
I have been trying to set up and run the PR in my local, but I couldn't. So instead of keep pining folks for help and delaying my comments, I will share feedback based on the screenshots shared. Sorry for jumping in late.

The page looks great across breakpoints, I just noticed the following in mobile:
1. Images in Features section are slightly bigger compared to the mockup.
2. Top and bottom paddings look tighter compared to the mockup.
3. Section headings, below Try WordPress action, look smaller. The mockup's style is `Heading 5`.

The font-size reduction in Try WordPress works well.

adamwoodnz:
@panchovm thanks for the feedback!

> 3\. Section headings, below Try WordPress action, look smaller. The mockup's style is `Heading 5`

We have a rule in the parent theme for all the redesigned sites which reduces the size of H5 headings from 26px to 20px on mobile: https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/theme.json#L334

Does this rule need to be adjusted? If it's not intended to be applied globally, we could specifically resize the other places it's applied, on the Download pages:

'Everything you need to set up your site just the way you want it.'


'Number of WordPress 6.1 downloads'


I have reduced the size of the Features icons to match the design (now 58px high), and increased the padding on the header and features sections as they seemed the furthest off. Again these paddings are controlled globally and clamped on mobile, so to some degree we try to go with these settings that have been developed on other pages.

fcoveram (approved):
The change looks great. We should keep consistency across the site, so your approach is perfect.

Thanks for the work ✨

PR #159, opened 2023-01-09
Try new design for search results page

ryelle:
This PR adds a new block for the Google Search embed & a new template for the Search page which uses that new block. The block itself is a very simple port of the code currently in [wporg-main/page-search.php](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-main/page-search.php).

This is a very shallow update of the page, really just changing the header font & colors. The search embed is provided by Google, so we probably can't style it very much - I don't think we can be guaranteed that the markup will stay consistent, so less customization is better.

See https://github.com/WordPress/wporg-documentation-2022/issues/30

### Screenshots

| Before | After |
|--------|-------|
| before | after |

1. View the `/search/` page
2. Try different search terms with this format `/search/block+editor`, or by typing in the search box on the page.
3. Try clicking the different tabs.
4. There should be no functional differences between production & this branch, only styles.

You can test this locally, kind of — it will only render the search box. It's best to push this up to a sandbox and test there.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/templates/page-search.html
+++ b/source/wp-content/themes/wporg-main-2022/templates/page-search.html
@@ -1,0 +1,27 @@
+<!-- wp:wporg/global-header /-->
+
+<!-- wp:group {"tagName":"main"} -->
+<main class="wp-block-group">
+	<!-- wp:cover {"overlayColor":"blueberry-1","minHeight":300,"align":"full"} -->
+	<div class="wp-block-cover alignfull" style="min-height:300px"><span aria-hidden="true" class="wp-block-cover__background has-blueberry-1-background-color has-background-dim-100 has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group"><!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"top":"0","right":"0","bottom":"0","left":"0"}}},"fontSize":"heading-cta"} -->
+	<h1 class="wp-block-heading has-text-align-center has-heading-cta-font-size" style="margin-top:0;margin-right:0;margin-bottom:0;margin-left:0">Search</h1>
+	<!-- /wp:heading -->
+	
+	<!-- wp:paragraph {"align":"center","className":"is-style-short-text"} -->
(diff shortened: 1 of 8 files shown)


source/wp-content/themes/wporg-main-2022/templates/page-search.html
```diff
@@ -0,0 +1,27 @@
+<!-- wp:wporg/global-header /-->
+
+<!-- wp:group {"tagName":"main"} -->
+<main class="wp-block-group">
+	<!-- wp:cover {"overlayColor":"blueberry-1","minHeight":300,"align":"full"} -->
+	<div class="wp-block-cover alignfull" style="min-height:300px"><span aria-hidden="true" class="wp-block-cover__background has-blueberry-1-background-color has-background-dim-100 has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group"><!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"top":"0","right":"0","bottom":"0","left":"0"}}},"fontSize":"heading-cta"} -->
+	<h1 class="wp-block-heading has-text-align-center has-heading-cta-font-size" style="margin-top:0;margin-right:0;margin-bottom:0;margin-left:0">Search</h1>
```
StevenDufresne, line 8: Does this need to be translatable?

StevenDufresne:
Assuming that JS/PHP was ported with minor changes, looks relatively straightforward. 
👍

source/wp-content/themes/wporg-main-2022/templates/page-search.html
ryelle: Yep, I'll swap this out for a pattern.

ryelle: Actually, we should probably use the syncing step since this is real page, but the page needs to be using a block theme template first, so I'll merge this and then set up the pattern with translations.

ryelle: Done in https://github.com/WordPress/wporg-main-2022/commit/e30c747d151bdcf1fa0947339a255783d54fd08b

PR #171, opened 2023-01-17
Add About landing page

### LINKED ISSUE #148: Add About landing page
Here's the [design](https://www.figma.com/file/W4jJEYbTBLAeMz5GzU1beL/About-Pages?node-id=1%3A2193&t=wTInpVpaFDxkPWG5-0):

| Desktop | Mobile |
|--------|-------|
| About Desktop | About Mobile |

This should be done as a template and pattern, similar to [Enterprise](https://github.com/WordPress/wporg-main-2022/pull/157)

adamwoodnz:
Closes #148

Props @tellyworth for initial content

Note that the design uses Courier Prime for the monospace serif font, but our theme has 'IBM Plex Mono', so that's what I'm using for now. I have [notified design](https://dotorgdesign.wordpress.com/2022/11/04/about-redesign/#comment-1727).

### Screenshots

| Desktop | Tablet | Mobile |
|--------|-------|-------|
| localhost_8888_about-2_(Desktop) (1) | localhost_8888_about-2_(iPad) (1) | localhost_8888_about-2_(Samsung Galaxy S20 Ultra) (3) |

1. Build the styles
2. Add a page at `/about-2`
3. Open the page, test responsive behaviour, links, cross browser layout. The cover in particular should fluidly scale between 600px and 1320px


StevenDufresne:
Nice! A couple things:

- The link arrows are still a problem for me since they are internal links. ([Issue](https://github.com/WordPress/wporg-parent-2021/issues/64))
- The header is black in the designs. Not sure if that was changed from the [figma file](https://www.figma.com/file/W4jJEYbTBLAeMz5GzU1beL/About-Pages?node-id=1%3A2193&t=MegGFDIyR2rSc8RG-0) I am looking at.
- Based on the same figma, the alignment of the "the four freedoms" content and the number list below is staggered. In this PR it's not. I like this PRs version better but that is a discrepancy i noticed.
- The space between the four freedoms content and the first freedom is a bit tight, can we make it larger? Not a strong opinion though...

| Designs | This PR |
| --- | --- |
| Screen Shot 2023-01-19 at 8 54 17 AM |Screen Shot 2023-01-19 at 8 42 44 AM |

adamwoodnz:
> * The header is black in the designs

Weird, the pattern builder generated the template as `black-on-white` even though the block in the prod page is `white-on-black`, nice catch 🙏

adamwoodnz:
> * Based on the same figma, the alignment of the "the four freedoms" content and the number list below is staggered. In this PR it's not. I like this PRs version better but that is a discrepancy i noticed.

Yeah good spotting. I found that at certain sizes the large numbers didn't fit alongside the wider columns so I decided to make them consistent with the other sections above (40/60)

adamwoodnz:
> * The space between the four freedoms content and the first freedom is a bit tight, can we make it larger?

Updated in https://github.com/WordPress/wporg-main-2022/pull/171/commits/8b36d9ec82a120e96a9305d4d22cccf72fc0b7ea

jasmussen:
Looks good, I like Plex over Courier, but I'll let Marko chime in once has a chance. 

We need to update the pencil strokes per some feedback:

screenshot-2023-01-18-at-13 47 16

The vectors are in [this figma](https://www.figma.com/file/W4jJEYbTBLAeMz5GzU1beL/About-Pages?node-id=1161%3A27048&t=Jgmws4t1IlVNc0IP-1), and can be exported as PNGs or SVGs, let me know what works best.

I'm not entirely sure about the pencil strokes further down the page, we can probably omit them from now and include only those at the top as a starting point. What do you think?

adamwoodnz:
> I'm not entirely sure about the pencil strokes further down the page, we can probably omit them from now and include only those at the top as a starting point. What do you think?

Yeah I'm not 100% sure about the positioning of them (mission swirl seems a little random, and four freedoms touches the text), but I quite like the idea of including them further down.

adamwoodnz:
> We need to update the pencil strokes per some feedback:

New screenshots @jasmussen

| Desktop | Tablet | Mobile |
|--------|-------|-------|
| localhost_8888_about-2_(Desktop) (2) | localhost_8888_about-2_(iPad) (2) | localhost_8888_about-2_(Samsung Galaxy S20 Ultra) |

jasmussen:
Looks good! Nice.

source/wp-content/themes/wporg-main-2022/patterns/about.php
```diff
@@ -0,0 +1,277 @@
+<?php
+/**
+ * Title: About 2
+ * Slug: wporg-main-2022/about-2
+ * Inserter: no
+ */
+
+?>
+<!-- wp:wporg/global-header /-->
```
ryelle: Does this create duplicate header & footer on the page? The header & footer are added in the page template too.

source/wp-content/themes/wporg-main-2022/src/style/style.scss
```diff
@@ -95,6 +95,131 @@ html[dir="rtl"] :where([style*="border-right-color"]) {
...
+	overflow-wrap: initial;
+}
+
+.wporg-about-section-software {
+	position: relative;
+}
+
+@media (min-width: 1200px) {
+	.wporg-about-cover-title {
+		margin-block-end: -42px !important;
```
ryelle: I know GB uses logical properties, but I don't think we've used them yet — should we be concerned about browser support?

ryelle: Could you add a note about where this number comes from?

source/wp-content/themes/wporg-main-2022/src/style/style.scss
```diff
@@ -95,6 +95,131 @@ html[dir="rtl"] :where([style*="border-right-color"]) {
...
+
+	[class*="wporg-about-section-freedom-"] {
+		margin-block-end: -30px !important;
+		position: relative;
+	}
+}
+
+@media (min-width: 600px) {
+	.wporg-about-cover-title {
+		margin-block-start: -30px !important;
```
ryelle: Can you use the spacing values for these numbers?

ryelle (commented):
Looks good - some questions about the code but nothing huge/blocking.

source/wp-content/themes/wporg-main-2022/patterns/about.php
adamwoodnz: Good catch, it doesn't seem to, but I'll remove it

adamwoodnz: Done in [75634c2](https://github.com/WordPress/wporg-main-2022/pull/171/commits/75634c24d19316184701248ea2c7a7804aa8d4a2)

source/wp-content/themes/wporg-main-2022/src/style/style.scss
adamwoodnz: > should we be concerned about browser support?

Good question! Looks ok https://caniuse.com/?search=margin-block-end

source/wp-content/themes/wporg-main-2022/src/style/style.scss
adamwoodnz: Done in [ea616fc](https://github.com/WordPress/wporg-main-2022/pull/171/commits/ea616fc17694ec7d61a8cc33142a7b7b34bbbe42)

source/wp-content/themes/wporg-main-2022/src/style/style.scss
adamwoodnz: > Could you add a note about where this number comes from?

Done in [ea616fc](https://github.com/WordPress/wporg-main-2022/pull/171/commits/ea616fc17694ec7d61a8cc33142a7b7b34bbbe42)

jasmussen:
Not something to block this PR, and can even happen post launch if need be, but there are some additional brush strokes down the page in some of the other sections in [the figma](https://www.figma.com/file/W4jJEYbTBLAeMz5GzU1beL/About-Pages?node-id=1161%3A27048&t=80MZrpJy5yOJoTsz-1):
Screenshot 2023-01-25 at 08 40 19

Screenshot 2023-01-25 at 08 40 22

adamwoodnz:
> some additional brush strokes down the page

💯 I can add these @jasmussen, sorry I thought they were still in consideration. Not sure if you saw my comment but I did [add my 2 cents above](https://github.com/WordPress/wporg-main-2022/pull/171#issuecomment-1399601374), as you asked 🙂

jasmussen:
Oh, apologies, I missed that. To emphasize, they aren't crucial for launch, and yeah I just got them sanity checked with a couple of others so they're good enough as is. I realize that if we're doing an editor only thing here, it might be hard to position them, and we can finesse that a little bit, it doesn't have to touch the text for example.

adamwoodnz:
Added the extra brushstrokes @jasmussen, they only show on screens >= 782px at present (phone/tablet landscape), but could make room on smaller devices if necessary. New screenshots:

| Desktop | Tablet landscape | Mobile |
|--------|-------|-------|
| localhost_8888_about-2_(Desktop) (3) | localhost_8888_about-2_(iPad) (5) | localhost_8888_about-2_(Samsung Galaxy S20 Ultra) (4) |

jasmussen:
Looks good, let's try this and revisit if need be.

marko-srb:
Hey everyone, 

Thank you for working on the About page diligently and informing us about the roadblocks and things that seem odd. :) 

Here is all context that I think you need for a few of the 'roadblocks' you experienced.

- Fonts. We should use **Courier Prime** here for the About page. It was one of the main design aspects when Pablo Honey and Beatriz Fialho were considering this approach. The about page should 'get off the rails' in a sense and be a different piece of content, connected through simplicity to a writing visual feel. Therefore, let me know if there are additional issues with implementing this. Thank you. 

- Brush strokes. In my absence, there was some updating with the brush strokes that @jasmussen already introduced. In [this Figma link](https://www.figma.com/file/W4jJEYbTBLAeMz5GzU1beL/About?node-id=1%3A2193&t=Q6Xit9C8dJND7PxI-1), you can find final look for these. 


Let me know if I missed anything or if you need any additional help. We can huddle on Slack if something is more complex. 

Thank you!

adamwoodnz:
@marko-srb 

> We should use **Courier Prime** here for the About page
> The about page should 'get off the rails' in a sense and be a different piece of content

So does this mean we don't see Courier being used on any other redesigned pages? Asking because it dictates whether I include it only in this repo or in a shared one

marko-srb:
> So does this mean we don't see Courier being used on any other redesigned pages? Asking because it dictates whether I include it only in this repo or in a shared one

That's correct @adamwoodnz. We should only use Courier Prime on About page. :)

adamwoodnz:
The new Typewriter block styles can't be applied in prod and added to the pattern yet, but they look like this:

Screen Shot 2023-01-31 at 3 32 10 PM
Screen Shot 2023-01-31 at 3 32 26 PM

Once merged I can use the block styles and update the pattern.

StevenDufresne:
The block styles implementation looks right 👍. 

I'm not certain about using block styles for typography though. I think it's more intuitive to use the font family dropdown. It's what you would do in most other text editors. Additionally, you can't combine them, so if a style comes around that does need a block style (maybe like a graphic styling), we'll need to expand the scope of the "typewriter" style, or rename it... 🤔 

I'm also torn as to whether we should keep all the fonts together in the wporg-mu-plugin folder as a convention. I get why it's added here but since our font _system_ is particular, it may make more sense for it to live where [we have documentation](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/bin/font-subset.js). Since the fonts are dynamically loaded, the only downside of having it in [wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/fonts) would be the ~40 extra lines in the [.css file](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/global-fonts/style.css). 

Thoughts?

adamwoodnz:
> I think it's more intuitive to use the font family dropdown

I think this is key. I guess I was trying to restrict the usage to certain blocks but that's less important.

> keep all the fonts together in the wporg-mu-plugin folder as a convention

If we are treating that as our font library rather than `global-fonts` as the name suggests, then I think you're right, if there is no load penalty.

source/wp-content/themes/wporg-main-2022/patterns/download.php
```diff
@@ -131,19 +131,19 @@
 <!-- wp:column -->
 <div class="wp-block-column"><!-- wp:list {"className":"is-style-links-list"} -->
 <ul class="is-style-links-list"><!-- wp:list-item -->
-<li><a href="https://learn.wordpress.org/course/getting-started-with-wordpress-get-setup/"><?php _e( 'WordPress courses ↗', 'wporg' ); ?></a></li>
+<li><?php _e( '<a href="https://learn.wordpress.org/course/getting-started-with-wordpress-get-setup/">WordPress courses ↗</a>', 'wporg' ); ?></li>
```
ryelle: I don't think these strings should be changing. I think it's a result of the parser changes? In any case, the link tag itself doesn't need to be translated.

I think it would be fine if you drop the changes to these other, non-about patterns, and then iterate on the parser along with the tests PR https://github.com/WordPress/wporg-main-2022/pull/181 so that these don't change.

adamwoodnz: Ah yes this was based on trunk not the parser branch, thanks for that. Now updated and I'll drop the changes to other files in that PR.

ryelle (dismissed):
Page and styles look good to me 👍🏻

adamwoodnz:
Added Courier Prime and brushstrokes @marko-srb



I'm going to merge and deploy the draft pages this afternoon and we can iterate more during final review phase.

PR #185, opened 2023-02-01
Add Courier Prime font option

### LINKED ISSUE #184: Add Courier Prime font to theme
Required for the redesigned About page. The font is now added to [wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/global-fonts/style.css#L2), but needs to be added as an option in this theme.

adamwoodnz:
Closes #184 

### Screenshots

| Editor | Frontend |
|--------|-------|
| Screen Shot 2023-02-01 at 4 25 25 PM | Screen Shot 2023-02-01 at 4 19 09 PM |

1. In the editor, edit a page using the new theme (main, download, etc) 
2. Select some text and Courier Prime should be available as a font option in the dropdown
3. Choose Courier Prime, update and preview the page
4. Courier Prime should be used on the frontend
5. Try changing the text to bold (italic is not supported)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2274,12 +2274,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/WordPress/wporg-mu-plugins.git",
-                "reference": "7d2ffaabca7d6c90c74190b09c23b76aa70ece58"
+                "reference": "2097028ce55d25b166ca8a234b0058359456ed7b"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/WordPress/wporg-mu-plugins/zipball/7d2ffaabca7d6c90c74190b09c23b76aa70ece58",
-                "reference": "7d2ffaabca7d6c90c74190b09c23b76aa70ece58",
+                "url": "https://api.github.com/repos/WordPress/wporg-mu-plugins/zipball/2097028ce55d25b166ca8a234b0058359456ed7b",
+                "reference": "2097028ce55d25b166ca8a234b0058359456ed7b",
                 "shasum": ""
             },
             "require-dev": {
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-main-2022/theme.json
```diff
@@ -1,7 +1,32 @@
...
+				{
+					"fontFamily": "'Inter', sans-serif",
+					"slug": "inter",
+					"name": "Inter"
+				},
+				{
+					"fontFamily": "'IBM Plex Mono', monospace",
+					"slug": "monospace",
+					"name": "Monospace"
+				}
```
adamwoodnz, line 26: @ryelle @StevenDufresne 

I expected only to have to add the new Courier Prime option (the others are already in the [parent theme.json](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/theme.json#L519)) and the entries would be merged, but unfortunately that's not the case and the list is overwritten. I haven't looked for a Gutenberg issue for this yet. Probably worth a comment here. 

Let me know if you know a workaround.

ryelle: It's the same root cause as this issue, array options overwrite the parent: https://github.com/WordPress/gutenberg/issues/40557. I haven't found a good workaround, especially not one that would work here :/

composer.lock
```diff
@@ -2274,12 +2274,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/WordPress/wporg-mu-plugins.git",
-                "reference": "7d2ffaabca7d6c90c74190b09c23b76aa70ece58"
+                "reference": "2097028ce55d25b166ca8a234b0058359456ed7b"
```
ryelle, line 2277: FWIW, whenever this is out of date, I just commit the change to trunk directly — it removes the diff noise from a PR, and it's probably necessary for anyone working on the project.

PR #186, opened 2023-02-01
Try a filter to add Courier Prime

ryelle:
See https://github.com/WordPress/wporg-main-2022/pull/185#discussion_r1092729807 — Rather than redefine all the fontFamilies in our child theme theme.json, I decided to try adding a filter to inject just the new font to the list. To do that, I filtered the user theme settings, which is something we're not using on the redesigned sites. I pulled out the existing list of fonts, then append the new font to that list - otherwise we get the same overwriting issue that the child theme's `theme.json` has.

This works, but I'm on the fence about whether this is a better idea (or being too clever). Maybe it's the start to a better idea?

### Screenshots

| Front end | Backend font dropdown |
|--------|-------|
| Screenshot 2023-02-01 at 1 14 39 PM | Screenshot 2023-02-01 at 1 15 06 PM |

Same as #185 — try using Courier Prime in different blocks. The other fonts (monospace, inter, etc) should still be available.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -20,6 +20,7 @@ require_once __DIR__ . '/src/release-tables/index.php';
 add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
 add_filter( 'wp_img_tag_add_loading_attr', __NAMESPACE__ . '\override_lazy_loading', 10, 2 );
+add_filter( 'wp_theme_json_data_user', __NAMESPACE__ . '\inject_theme_json' );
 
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
Thanks for trying this. Tested and works for me. Seems an effective workaround while we wait for an API which supports merging. I'm trying to think how it could go wrong (in terms of being too clever)... it doesn't appear to affect other `settings` entries being added to the child theme.json, so seems ok to me.

Shame we can't add comments to JSON to mention the filter. Could go in the README if we think it's significant enough.

@StevenDufresne thoughts?

StevenDufresne:
I'm good with it. 

I had to do similar things with wp-themes.com to get [style variations to work](https://github.com/WordPress/wordpress.org/blob/trunk/wp-themes.com/public_html/wp-content/plugins/style-variations/inc/page-intercept.php#L95). It seems to be stable enough.

PR #190, opened 2023-02-02
Update About sub pages

### LINKED ISSUE #189: Update About sub page list item spacing and content
1. List items with long content should have 20px spacing
2. Old WordPress logo should be removed from About/Security page

adamwoodnz:
Fixes #189 

List items with long content have more space between, and the old logo is removed from About/Security

### Screenshots

| Before | After |
|--------|-------|
| image | Screen Shot 2023-02-03 at 9 57 01 AM |

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/features.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/features.php
@@ -6,8 +6,8 @@
  */
 
 ?>
-<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"right":"var:preset|spacing|60","left":"var:preset|spacing|60","top":"16px","bottom":"16px"}}},"backgroundColor":"white","className":"is-sticky","layout":{"type":"constrained"}} -->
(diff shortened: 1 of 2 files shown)


source/wp-content/themes/wporg-main-2022/patterns/features.php
```diff
@@ -107,16 +107,16 @@
 <!-- /wp:list-item --></ul>
 <!-- /wp:list -->
 
-<!-- wp:heading -->
+<!-- wp:heading {"className":"wp-block-heading"} -->
 <h2 class="wp-block-heading"><?php _e( 'Developer Features', 'wporg' ); ?></h2>
```
ryelle, line 111: Not a blocker, but do you know why this className is being added? I thought that was a core class for the heading block.

ryelle (approved):
Lists look good, and image is gone from Security 👍🏻

source/wp-content/themes/wporg-main-2022/patterns/features.php
adamwoodnz: Yeah I'm not sure, seems to be being added to all the headings each time I update the patterns. I assumed it was meant to be there and guessed that maybe it was dropped by a bug at some point, and the last core update added it back?

PR #194, opened 2023-02-03
Fix permission check for superadmins

tellyworth:
Fixes a bug where superadmin users can't save new drafts.

`current_user_can()` always returns true for superadmin users, so this is a more accurate way of checking for a role.


adamwoodnz (approved):
Works for my super admin user in sandbox

Issue #219, opened 2023-03-23
Download modal layout is broken


adamwoodnz:
**Describe the bug**

![image](https://user-images.githubusercontent.com/1017872/227083220-8c4b3dfd-9da3-4ae7-aa77-94922c7dfad2.png)

**To Reproduce**
1. Go to https://wordpress.org/download/
2. Click on 'Get WordPress 6.x.x'
4. See that the header layout is broken

**Expected behavior**
Should look more like this with one solid box of colour

![image](https://user-images.githubusercontent.com/1017872/227083898-bdad1846-5c04-4c21-ac96-79888bfc8a40.png)

PR #224, opened 2023-03-27
Fix deployment of styles of Download modal without other block changes

adamwoodnz:
Change version of Download Modal style to the filemtime of that file.

If it's set to the time of the index.asset.php file and that file hasn't changed, the old version is loaded

Fixes the deployment of #219

Tested in sandbox, where the new version number is `1679876118`. Updating the CSS request in prod with the same version loads the new CSS:

Screen Shot 2023-03-27 at 1 38 09 PM


StevenDufresne (approved):
👍 LGTM

PR #248, opened 2023-04-04
Search: Update search page styles

### LINKED ISSUE #216: Search page: Basic improvements.
Let's improve the [search page](https://wordpress.org/search/jetpack/):

Screenshot 2023-03-21 at 10 51 47 PM

* [ ] The search header does not need to be 300px high, and doesn't need the subtext. Let's try a blue breadcrumb bar with just the "Search" subheading.
* [ ] Omit the register / login bar. In general, let's omit this on every single page we can, and show it only in the cases where it's relevant. On the pages where we do include the login bar, let's use the same bg color for the login bar, as the main navigation bar background, or integrate the login link in the prose.
* [ ] "Learn" and  "Forums" highlighted in menu, which doesn't seem relevant to search. Let's not highlight anything.

It appears that the search results use an embedded Google search. This comes with a couple of issues, design-wise:

* It becomes a separate search box instead of being unified with what exists in the header. Main problem with that is the disconnect with the search box in the header, which even if it kicked off the search, on the results page becomes inert and collapsed. There are a few ways we an thread this, designwise — "moving" the search box, unifying on the one in the header with it staying open, etc.  
* This duplicated search box is awkwardly balanced, low contrast, somewhat small in terms of emphasis (It's the most important part of the page).
* The search icon is reversed and different from the one in the header.
* The search tabs provided are not in sync with the header verbiage ("Support Forums" instead of "Forums", "Support Docs" instead of "Documentation", separate "Developer Docs", etc)
* The formatting of the results leaves a great deal to be desired. Depending on what's possible here, I'd love to provide a mockup.

What options do we have for designing the Google search results? And if our options are limited, are there alternatives we can explore?

ryelle:
In this PR, I've updated the page content to use the local nav style for the "Search" header, and added some CSS. It should flip the search icon, though I didn't go as far as overwriting it in case the embed code changes one day; update the light grey color of the results count to be `charcoal-4` and force everything to use the theme's body font (Inter).

I tried updating the style provided by Google on the search form & results to use our color palette - I could only change some things (text colors, the basic button colors, and the search box's border), but I think it helped increase the contrast on the page. These changes to the embed are published to production already.

Fixes #216

### Screenshots

| Before | After |
|--------|-------|
|  |  |

1. Go to the search page `/search/`
2. Search for something
3. The results should look like the after screenshot

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/search.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/search.php
@@ -6,20 +6,16 @@
  */
 
 ?>
-<!-- wp:cover {"overlayColor":"blueberry-1","minHeight":300,"align":"full"} -->
-<div class="wp-block-cover alignfull" style="min-height:300px"><span aria-hidden="true" class="wp-block-cover__background has-blueberry-1-background-color has-background-dim-100 has-background-dim"></span><div class="wp-block-cover__inner-container"><!-- wp:group {"layout":{"type":"constrained"}} -->
-<div class="wp-block-group"><!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"top":"0","right":"0","bottom":"0","left":"0"}}},"fontSize":"heading-cta"} -->
-<h1 class="wp-block-heading has-text-align-center has-heading-cta-font-size" style="margin-top:0;margin-right:0;margin-bottom:0;margin-left:0"><?php _e( 'Search', 'wporg' ); ?></h1>
-<!-- /wp:heading -->
-
-<!-- wp:paragraph {"align":"center","className":"is-style-short-text"} -->
-<p class="has-text-align-center is-style-short-text"><?php _e( 'Search the WordPress.org website for plugins, themes, and support.', 'wporg' ); ?></p>
+<!-- wp:group {"align":"full","style":{"elements":{"link":{"color":{"text":"var:preset|color|white"}}},"spacing":{"padding":{"top":"16px","right":"var:preset|spacing|60","bottom":"0","left":"var:preset|spacing|60"}}},"backgroundColor":"blueberry-1","textColor":"white","className":"is-style-brush-stroke is-sticky","layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull is-style-brush-stroke is-sticky has-white-color has-blueberry-1-background-color has-text-color has-background has-link-color" style="padding-top:16px;padding-right:var(--wp--preset--spacing--60);padding-bottom:0;padding-left:var(--wp--preset--spacing--60)"><!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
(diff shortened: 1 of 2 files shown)


StevenDufresne:
I tested these changes, nice improvement! Although I don't think it's included in the original issue, but I think we could probably bump up the pagination font-size and maybe add more padding around it. It's pretty small. That's an opinion. 

I'm good with this being merged as is.

outdoor2kode (approved):
The changes are tested and look good 👍

jasmussen:
Thanks, this is a vast improvement. I still think we need to revisit this page in lieu of how much or how little we can style the results. From a little inspecting, it seems like the results are not in an iframe after all, so it should be possible to create enough CSS to override and make things look right if need be. But this would be a hacky endeavour, and subject to breakage if the upstream code changed. I'll refer to technical advice on the feasibility of this, but if it would be useful, I would be happy to provide some CSS to shore things up here. 

Things that come to mind:

* Improve the apperance of the tabs in terms of font size, active tab appearance, border color and position.
* Improve the apperance of the "Relevance" button.
* Improve the appearance of the search field — not so wide, 2px rounded corners, integrated search button with the right icon, hide or improve the X to clear icon.
* Bring better spacing and emphasis to pagination at the bottom, make it match pagination in /news ideally.

Also, hide, or correctly scale the "Enhanced by Google" background that appears in the search field in some cases.
Screenshot 2023-04-11 at 11 18 24

There's something wrong with the auto complete, it covers results:

auto complete


Not sure if it's possible, but if it is, even now we can rename the tabs to be more appropriate:

* Instead of "Support Forums", it should be "Forums" to match the menu item in Learn.
* Instead of "Developer Documentation", it should be "Developers" to match the menu item in Learn.
* Instead of "Support Docs", it should be "Documentation" to match the menu item.

There's a lot more and I'm happy to collect it all on a new issue if need be, but much of it is dependant on whether we continue with Google search or not.

ryelle:
>  But this would be a hacky endeavour, and subject to breakage if the upstream code changed. I'll refer to technical advice on the feasibility of this, but if it would be useful, I would be happy to provide some CSS to shore things up here.

I did add a few minor CSS overrides but I'd like to keep that to a minimum in case future changes break the page. A few updates to the active tabs & pagination would probably be fine, but I imagine you want larger changes to the "Relevance" dropdown & search button that might break them if the markup changes. That said, feel free to make a PR yourself if you want to suggest CSS.

> Also, hide, or correctly scale the "Enhanced by Google" background that appears in the search field in some cases.

I was able to disable this in the search settings 👍🏻 

> There's something wrong with the auto complete, it covers results:

Ah, the header offset is messing with the positioning, I should be able to fix that.

> Not sure if it's possible, but if it is, even now we can rename the tabs to be more appropriate:

No, I [mentioned on the issue itself](https://github.com/WordPress/wporg-main-2022/issues/216#issuecomment-1496633800), we can't change those.

> There's a lot more and I'm happy to collect it all on a new issue if need be, but much of it is dependant on whether we continue with Google search or not.

I don't know if there's a concrete path forward yet, but I'm pretty sure we'll move away from Google here.

jasmussen:
Thank you!

PR #250, opened 2023-04-10
Download page: Update download button styles on small screens

### LINKED ISSUE #249: Download: Change download section on mobile to emphasize installation guide
On a physical mobile device, downloading the zip file should certainly be possible, but it probably shouldn't be the emphasized action. Here's a design that changes this:

Download

What do you think?

ryelle:
Fixes #249 — Update the style of the "Download" and "Installation guide" links when viewing from a small screen. The cutoff for "small" is 600px.

### Screenshots

| Before | After |
|--------|-------|
| trunk | branch |

1. Check out this branch & build the CSS
2. View the download page
3. On larger (600+) screens, the Download button is primary
4. On smaller screens, the Installation guide button is primary

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/src/style/style.scss
+++ b/source/wp-content/themes/wporg-main-2022/src/style/style.scss
@@ -427,6 +427,42 @@ html[dir="rtl"] :where([style*="border-right-color"]) {
 		width: 100% !important;
 	}
 
+	#download-install .wp-block-button:nth-child(1) a {
+		padding: 0;
+		background-color: transparent;
+		color: var(--wp--custom--link--color--text);
+		font-weight: 400;
+		text-align: revert;
+		display: inline;
+
+		&:hover,
+		&:focus-visible {
+			text-decoration: underline;
+		}
+	}
+
+	#download-install .wp-block-button:nth-child(2) a {
+		--wp--custom--button--outline--color--background: var(--wp--preset--color--blueberry-1);
+		--wp--custom--button--outline--color--text: var(--wp--preset--color--white);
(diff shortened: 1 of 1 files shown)


StevenDufresne (commented):
Tested and the code makes sense. 👍 

Noting that clicking the Installation button on mobile brings us to the documentation page that has an [overflow problem](https://github.com/WordPress/wporg-documentation-2022/issues/50#issuecomment-1502658546). We should probably consider fixing it in this iteration as we'll send more people down that path.

jasmussen:
Nice and simple, and a good win for mobile devices. 

Not blocking, but mostly for discussion. Instead of using a width query for this, should we use something like this instead?

```
@media (hover: none) {
	...
}
```

Admittedly super edge-casey, but it doesn't seem useful for an iPad either to download a zip file (though I'm aware things have changed here in recent years so I might be wrong there).

ryelle:
> Instead of using a width query for this, should we use something like this instead?

I think it's more of the same, there's always an edge case. People do work from tablets, there are IDEs and FTP apps for iOS and Android. It's also not clear what's defined as the "primary" pointing device, so would a touchscreen laptop have `hover:none`? I wouldn't think so, but [at least one convertible laptop reports hover:none](https://andreasburg.de/browser-hover-check.html). We could use `any-hover: none` to see if the client has any pointer device, but would still capture regular tablet users.

On the other hand, with screen size, someone could just have a small browser window on a desktop. Zooming a screen can also trigger the width media query at larger viewport sizes on some browsers (Safari for example).

Trying to nail down this heuristic is one part of why I'm not a fan of switching content for different devices. I don't know which edge cases are better, so if you want this changed later we can open a new issue for it.

fcoveram:
This change looks great to me. Not sure if I should approve the PR or not

PR #251, opened 2023-04-11
Adjust heading sizes and margin

### LINKED ISSUE #246: About/Philosophy: Adjust heading sizes
Filing here as it's a general issue about heading sizes, though at the moment this is most visible on the About subpages. Specifically, the heading sizes on https://wordpress.org/about/philosophy/ are a bit off. While we can do a larger design pass separately, it would be good to address the heading hierarchy with a quick fix in the mean time. Suggestion:

For starters, let's use the full 70px heading sizes for the main headings (same as /news). We also need more visible spacing between the major sections. Currently the the top-margin on headings is 20px, inherited from WordPress. The above mockup shows that chnaged to 50px top margin on headings.

Apply the same across the other subpages, notably Graphics & Logos which has the same spacing issue.

CC: @panchovm in case you have some visual thoughts. The alternative to the 70px major heading would be smaller subheadings.

outdoor2kode:
Fixes #246 

This PR 
* Removes `has-heading-2-font-size` for `<h1>`. Except for the `<h1>` in the download subpage. Reason see below.
* Adds `margin-top:var(--wp--preset--spacing--40)` for `<h2>` and `margin-top:var(--wp--preset--spacing--30)` for `<h3>`. And the `<h2>` and `<h3>` that have their font size overridden weren't added with it, as the layout would go off consequently.

### Philosophy
| Before | After |
|--------|-------|
| Screen Shot 2023-04-11 at 10 54 26 AM | Screen Shot 2023-04-11 at 10 53 00 AM |

### Graphics & Logos
| Before | After |
|--------|-------|
| Screen Shot 2023-04-11 at 11 01 08 AM | Screen Shot 2023-04-11 at 11 02 28 AM |

### Security
| Before | After |
|--------|-------|
| Screen Shot 2023-04-11 at 11 01 15 AM | Screen Shot 2023-04-11 at 11 02 47 AM |
| Screen Shot 2023-04-11 at 11 01 27 AM | Screen Shot 2023-04-11 at 11 02 54 AM |

### Download Subpage

The adjustments made to the font size of the `<h1>` on the download subpages were reverted as the layout was off, even with `nowrap` added. 

| Before | After | Nowrap |
|--------|--------|-------|
| Screen Shot 2023-04-11 at 11 13 26 AM | Screen Shot 2023-04-11 at 10 49 52 AM | Screen Shot 2023-04-11 at 10 49 58 AM |

1. Go to `/about/*`.
2. Check if `<h1>` is of `font-size` `70px` and `<h2>, <h3>` is of `margin-top: 40px, margin-top: 30px` respectively.

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/accessibility.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/accessibility.php
@@ -17,8 +17,8 @@
 <!-- /wp:group -->
 
 <!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","right":"var:preset|spacing|edge-space","left":"var:preset|spacing|edge-space","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--edge-space)"><!-- wp:heading {"level":1,"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}},"fontSize":"heading-2"} -->
-<h1 class="wp-block-heading has-heading-2-font-size" style="margin-bottom:var(--wp--preset--spacing--30)"><?php _e( 'Accessibility', 'wporg' ); ?></h1>
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--edge-space)"><!-- wp:heading {"level":1,"style":{"spacing":{"margin":{"bottom":"var:preset|spacing|30"}}}} -->
+<h1 class="wp-block-heading" style="margin-bottom:var(--wp--preset--spacing--30)"><?php _e( 'Accessibility', 'wporg' ); ?></h1>
 <!-- /wp:heading -->
 
 <!-- wp:paragraph -->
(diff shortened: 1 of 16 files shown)


jasmussen:
This seems a big win, let's land this and roll it out as fast as possible, thank you.

Not something that should block this from landing, but _potentially_ result in a polish followup if need be, I'd love a sanity check from @WordPress/meta-design on the spacings. It feels like we could have a bit more space _below paragraphs_. What do you think?

ryelle:
@renintw The page content needs to be updated on the production site & synced to the patterns with `yarn build:patterns` — otherwise these changes will be overwritten the next time the content is synced.

I tried a different approach in https://github.com/WordPress/wporg-parent-2021/pull/83 where the parent theme provides these margins, since the same values are used in Documentation. If that should not cascade from the parent theme, we can use this in-content margin approach.

If we go with the parent theme approach, the content still needs to be updated to remove the font size from the first heading.

> It feels like we could have a bit more space below paragraphs. What do you think?

For reference, the current block spacing is 20px. If you want to change it, can you open an new issue, and note if the change is for all blocks or just paragraphs, or only on specific pages or sites? So we can know if it should be a change at the page level, child theme, or parent theme. Also, feel free to open new issues for these kinds of questions, just tag it  "Needs Design Feedback" if it's just a designer discussion :)

jasmussen:
Let's keep as is and revisit.

fcoveram:
Agree. Keep it and revisit it later.

outdoor2kode:
> The page content needs to be updated on the production site & synced to the patterns with yarn build:patterns — otherwise these changes will be overwritten the next time the content is synced.

@ryelle Thanks 👍 I've run `build:patterns` and updated in [fd55f13](https://github.com/WordPress/wporg-main-2022/pull/251/commits/fd55f139efb9e11ff2f30a293340cd714e7b33f5).

**Update**
I found that the font size of the heading on production wasn't changed to `70px`, so I also used the edit site to edit the page template. I believe I should clear the customizations I made here and let the deployment afterwards change the template, right?

ryelle:
>  I found that the font size of the heading on production wasn't changed to 70px, so I also used the edit site to edit the page template. I believe I should clear the customizations I made here and let the deployment afterwards change the template, right?

That's correct, production isn't updated until the patterns are synced & deployed. I've removed the page templates you created since they were preventing the patterns from being used.

ryelle (approved):

PR #265, opened 2023-05-11
Patterns: Update the WP20 banner, use Link Wrapper block

### LINKED ISSUE #258: Improve the swag banner
Extracting from #253, and expanding with a little additional context or screenshots. 

> Swag Banner needs to be clickable as a full module.

That's this one:

Screenshot 2023-05-01 at 14 19 23

I'm not sure the technically best way to accomplish this given it's in the block editor, but one way to do it would be to link the image to the mercantile site, and then change the heading + paragraph + link into one big link, with large text and curated underline inside.
Another approach could be to add a link with a custom CSS class, then use custom CSS with the text-indent hack (`text-indent: 9999px`) to hide the text visually, then abs-position that link to cover the entire box. Accessibility wise, you'd just want to author the link with descritpively useful text, then omit the actual hyperlink from the "mercantile" spelled out link in the text that follows. Custom CSS should be available in Global Styles now, so you could create a helper class to handle it. 

>  Right now the only clickable element in a URL that reflects as a button/link when hovered. We need to add a clearer button or link.

The most basic solution I can think of is to just underline the link always:
Screenshot 2023-05-01 at 14 27 31

Another option would be rephrasing the text a little bit, and making the link into a button:
Screenshot 2023-05-01 at 14 29 37

Note that we'd still want the whole module clickable.

> The products in the Swag Banner are laid out strangely in scale and proportion, also the shadows can be better and consistent.

> In smaller desktop viewports, the banner pushes everything down too much. Could be less tall. In mobile it's even more obvious.

> The copy say Show your wp20 excitement, but it doesn't explain it's the anniversary. Without making the copy too long, it'd be good to bring slightly better awareness of it. Text and call to action needs to be more specific, even for unawareness. Celebrate the 20th anniverssary with new swag. Show off WordPress 20th with fresh swag. Or alike.

This one is a bit tricky since the media & Text banner needs to scale to mobile. But here's a tweak that rearranges things a bit, rephrases, makes it less tall, and tries to tie it all together. Desktop:

Alternative

* Facing, simpler illustration
* Stacking so that shadows _shouldn't_ be needed but let me know if I got this wrong
* Simpler, smaller text, more context, 240px tall.

Mobile:

Alternative mobile

* Should still be feasible to build this using the Media & Text block
* About 370px tall

It's likely going to be a challenge to make this substantially smaller, due to how the responsiveness is built-in to the block. If we need a different approach here, let me know any technical ideas you might have.

ryelle:
This PR switches the large banner on the homepage to a smaller banner, wrapped in a link. This also updates the content of the banner to highlight the WP20 website, instead of the swag.

Fixes #258, See https://github.com/WordPress/wporg-mu-plugins/pull/398 (required before this can be merged)

cc @DanSoschin

### Screenshots

| Size | Screenshot |
|--------|-------|
| 400px |  |
| 600px |  |
| 960px |  |
| 1400px |  |
| Focus state | focus |
| Hover state | hover |

1. Make sure https://github.com/WordPress/wporg-mu-plugins/pull/398 is applied or already merged
2. Check out this PR, build the CSS
3. View the homepage, it should have the smaller banner.


adamwoodnz (commented):
Works for me but doesn't quite match design:
1. Height in design is 60px
2. Confetti looks like it's lower opacity (can't see any full color orange or white discs)
3. The 'View Events' link is missing (maybe intentional due to the whole banner being clickable?)

Screen Shot 2023-05-12 at 10 20 54 AM

Screen Shot 2023-05-12 at 10 25 54 AM

Screen Shot 2023-05-12 at 10 20 44 AM

ryelle:
Most of that was intentional, but I'll let @marko-srb chime in if I should change anything

> Height in design is 60px

This size worked better on smaller screens, otherwise it felt like there wasn't enough space when the text wrapped to additional lines.




> Confetti looks like it's lower opacity (can't see any full color orange or white discs)

I put the overlay up to 50% based on feedback from @annezazu, "the dots in the background image make it hard to read the text" — maybe a different background image designed for this size would be better?

> The 'View Events' link is missing (maybe intentional due to the whole banner being clickable?)

Yeah, intentional because the whole line is a link now.

marko-srb (commented):
@ryelle 

All the updates are fine. One suggestion/question.

- Can we reduce text size on mobile only? (in original design I went with 13 for mobile)

If not, can we reduce the space from logo to the text, to about half of what's done. See screenshot:
Screen Shot 2023-05-12 at 15 57 22

Other than that, we can push this change.

P.S. It is fine to narrow that space on both responsive and desktop.

Thanks!

ryelle:
I've tweaked the spacing & updated the font size on small (<500px) screens:

| Large | Small |
|---|---|
| large screen | small screen |

PR #269, opened 2023-06-14
Try new Local Navigation block

ryelle:
This is a companion PR to https://github.com/WordPress/wporg-mu-plugins/pull/404, which introduces a new Local Navigation container to hold the section title & section menu. This PR extracts out the local header from the synced post content, and creates new reusable pattern-partials which handle the header content on each subpage. The new patterns use the Local Navigation container block.

The changes here also bring these headers in line with the generic section header mocked up. @WordPress/meta-design As I understood it, this pattern is complete, but if that's not the case let me know.

Section Header L2

❗ If we decide not to merge the Local Nav block, I still think we should merge https://github.com/WordPress/wporg-main-2022/commit/e1e42a3e60411e69d1f81841cb1ef0b61f31691c to DRY up the About pages. That won't change any visuals.

Additionally, once this is merged, I'll need to go in and update each of the subpages to remove the banner, so that the content sync doesn't overwrite these changes.

See WordPress/wporg-mu-plugins#465

### Screenshots

1. The Stats page, a general About subpage. I added borders to the header as this was mocked up at some point, but I don't know if that's still the preference.
2. The Requirements page, scrolled down— the top banner sticks under the global header, but the breadcrumbs scroll. Is that the expected behavior?
3. The Beta/Nightly page— I kept the brush stroke style on the download subpages, but that + the breadcrumbs is a little strange.
4. The download counter— unlike the other download subpages, this does not have the breadcrumbs, and uses the black+borders style for the banner. So we can still have unique features on individual pages.

| Before | After |
|--------|-------|
| before-about-stats | after-about-stats |
| before-about-reqs-scroll | after-about-reqs-scroll |
| before-download-nightly | after-download-nightly |
| before-download-counter | after-download-counter |

1. Make sure you have https://github.com/WordPress/wporg-mu-plugins/pull/404 applied on wporg-mu-plugins
2. Check out this branch
3. View the About and Download subpages

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/functions.php
+++ b/source/wp-content/themes/wporg-main-2022/functions.php
@@ -21,6 +21,7 @@ add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_action( 'init', __NAMESPACE__ . '\register_shortcodes' );
 add_filter( 'wp_img_tag_add_loading_attr', __NAMESPACE__ . '\override_lazy_loading', 10, 2 );
 add_filter( 'wporg_block_site_breadcrumbs', __NAMESPACE__ . '\update_site_breadcrumbs' );
+add_filter( 'render_block_core/site-title', __NAMESPACE__ . '\use_parent_page_title', 10, 3 );
 
 /**
  * Enqueue scripts and styles.
@@ -117,7 +118,7 @@ function override_lazy_loading( $value, $image ) {
 }
 
 /**
- * In the subpages, update first breadcrumb to be the parent page, rather than site home.
+ * Use the page heirarchy to display breadcrumbs.
  */
 function update_site_breadcrumbs( $breadcrumbs ) {
 	$parent = get_post_parent();
@@ -125,21 +126,51 @@ function update_site_breadcrumbs( $breadcrumbs ) {
 		return $breadcrumbs;
 	}
 
-	if ( 'about' === $parent->post_name ) {
-		$top_level_page = array(
-			'url'   => home_url( '/about/' ),
-			'title' => __( 'About', 'wporg' ),
+	$breadcrumbs = array();
+	$breadcrumbs[] = array(
+		'url' => false,
+		'title' => get_the_title(),
+	);
+
+	while ( $parent ) {
+		$breadcrumbs[] = array(
+			'url' => get_permalink( $parent->ID ),
+			'title' => $parent->post_title,
 		);
-		$breadcrumbs[0] = $top_level_page;
-	} else if ( 'download' === $parent->post_name ) {
-		$top_level_page = array(
-			'url'   => home_url( '/download/' ),
-			'title' => __( 'Download', 'wporg' ),
-		);
-		$breadcrumbs[0] = $top_level_page;
+		$parent = get_post_parent( $parent );
 	}
 
-	return $breadcrumbs;
+	return array_reverse( $breadcrumbs );
+}
+
+/**
+ * Replace the site title & link with the parent page title.
+ *
+ * The About and Download sections are pseudo-individual sites, so when site title
+ * is used there, it should reference the parent page.
+ *
+ * @param string   $block_content The block content.
+ * @param array    $block         The full block, including name and attributes.
+ * @param WP_Block $instance      The block instance.
+ */
+function use_parent_page_title( $block_content, $block, $instance ) {
+	$parent = get_post_parent();
+	if ( ! $parent ) {
+		return $block_content;
+	}
+
+	// Loop up to the first child page, this is the section title.
+	while ( $parent ) {
+		$url = get_permalink( $parent->ID );
+		$title = $parent->post_title;
+		$parent = get_post_parent( $parent );
+	}
+
+	return str_replace(
+		array( home_url(), get_bloginfo( 'name' ) ),
+		array( $url, $title ),
+		$block_content
+	);
 }
 
 /**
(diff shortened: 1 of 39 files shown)


jasmussen:
Thanks a bunch, this looks great.

I don't know for sure that we will _always_ have the borders visible, and depending on context (bg color etc), they might need to be inverted or semi-transparent. I'm not sharing that as blocking feedback, it's clear you're already handling on a case by case basis with the Download section. But it is to say, I would love to get some feedback from @panchovm when he gets a moment, notably around this one:



For example, Should it say "About WordPress.org" instead of "About" or "WordPress.org"? And should the submenu item match? I think Francisco may be back from some AFK tomorrow and can sanity check.

ryelle:
@jasmussen @fcoveram Do you think this can be merged, or do you have more feedback? The Local Navigation container block itself has already been merged, so the main issue here is whether the design works.

jasmussen:
I'm happy to iterate this, we can always revisit. I'll defer to Francisco in case he chimes in.

fcoveram:
Sorry for jumping in late here.

I checked the images and realized we did not explain the header correctly. I took some pages of Learn site to show how the header behaves at each depth level. See the diagram below

IA map and Wireframe with three header variants side by side

Per level depth:
* At level 1, the hero header shows all level 2 pages inactive
* At level 2, header shows the subsite name on the left and level 2 pages on the right. The page where you are is shown active in the menu
* At level 3, header shows the subsite name on the left and level 2 pages on the right. There is no active page and breadcrumb is visible. As pointed out [here](https://github.com/WordPress/wporg-mu-plugins/issues/465) (one that I couldn't find), "Home" is an idea for the name duplication that takes you to the subsite homepage (to "Learn", following the example).

Showing the breadcrumb at level 2 is redundant as it will show the home and active page links.

The border dividing header and breadcrumb is not always visible but depends on the subsite style. But as Joen said, it shouldn't block the ticket.

ryelle:
Thanks for the explanation! I'm pretty sure this PR only changes the "level 3" pages, so I think I've got it right? With the exception of the Download Counter. 

- L1: [WordPress.org homepage](https://wordpress.org/) (unchanged)
    - L2: [About](https://wordpress.org/about/) (unchanged)
        - L3: [About > Accessibility](https://wordpress.org/about/accessibility/)
        - L3: [About > Domains](https://wordpress.org/about/domains/)
        - L3: [About > Etiquette](https://wordpress.org/about/etiquette/)
        - L3: [About > Features](https://wordpress.org/about/features/)
        - L3: [About > History](https://wordpress.org/about/history/)
        - L3: [About > License](https://wordpress.org/about/license/)
        - L3: [About > Logos](https://wordpress.org/about/logos/)
        - L3: [About > Philosophy](https://wordpress.org/about/philosophy/)
        - L3: [About > Privacy](https://wordpress.org/about/privacy/)
        - L3: [About > Requirements](https://wordpress.org/about/requirements/)
        - L3: [About > Roadmap](https://wordpress.org/about/roadmap/)
        - L3: [About > Security](https://wordpress.org/about/security/)
        - L3: [About > Stats](https://wordpress.org/about/stats/)
    - L2: [Download](https://wordpress.org/download/) (unchanged)
        - L3: [Download > Beta/Nightly](https://wordpress.org/download/beta-nightly/)
        - L3: [Download > Counter](https://wordpress.org/download/counter/)
        - L3: [Download > Releases](https://wordpress.org/download/releases/)
        - L3: [Download > Source Code](https://wordpress.org/download/source/)

So does the breadcrumb need to update to be Home / About / [current page]? Or was there other feedback about these specific pages?

fcoveram:
I missed a crucial point in my explanation, [WordPress.org homepage](https://wordpress.org/) is not counted as L1. In the example attached, the subsite's homepage ([Learn homepage](https://learn.wordpress.org/)) is L1.

So following the site structure you shared above, the logic I suggested goes as follows:

* Root: [WordPress.org homepage](https://wordpress.org/)
   * L1: [About](https://wordpress.org/about/)
      * L2: [About > Accessibility](https://wordpress.org/about/accessibility/)
      * L2: [About > Domains](https://wordpress.org/about/domains/)
      * ...

What's currently in production works well because L1 (About) and L2 (Requirements, Features, Security, etc) don't show the breadcrumb but active style in the subnavigation area.

> ...As pointed out https://github.com/WordPress/wporg-mu-plugins/issues/465, "Home" is an idea for the name duplication that takes you to the subsite homepage (to "Learn", following the example).

In the "About" case, my comment quoted above means that "home" link in breadcrumb takes you to About's [homepage](https://wordpress.org/about/).

ryelle:
> [WordPress.org homepage](https://wordpress.org/) is not counted as L1. In the example attached, the subsite's homepage ([Learn homepage](https://learn.wordpress.org/)) is L1.

Ah right— we got caught up in a technical-vs-conceptual issue 🙂  technically, wordpress.org and learn.wordpress.org are the same level in each site (both roots/homepages), and About is a subpage in WordPress.org— but conceptually, you're thinking of About and Download as unique "sites" … or maybe it's more correct to say that the WordPress.org homepage itself is the special case? So would a page like [Enterprise](https://wordpress.org/enterprise/) or [Mobile](https://wordpress.org/mobile/) be L1, even though they have no child pages?

In any case, the updated PR now only shows the section title (About or Download) and the navigation, for each page. Like this:

Screen Shot 2023-07-07 at 11 46 07

adamwoodnz (approved):
👍 LGTM

fcoveram:
To @ryelle 

> …we got caught up in a technical-vs-conceptual issue 🙂 technically

> …you're thinking of About and Download as unique "sites"

> So would a page like [Enterprise](https://wordpress.org/enterprise/) or [Mobile](https://wordpress.org/mobile/) be L1, even though they have no child pages?

Yes, yes, and yes 😅 . I understand it's technically wrong, but the approach was taken for browsing purposes.

> the updated PR now only shows the section title (About or Download) and the navigation, for each page.

And it looks great ✨ 🚀

fcoveram (approved):

PR #273, opened 2023-06-27
Bump the recommended MariaDB version to 10.4

sabernhardt:
Matches the version number in the WordPress readme and Site Health.

https://core.trac.wordpress.org/ticket/58158
https://meta.trac.wordpress.org/ticket/7099


sabernhardt:
Thanks!

PR #275, opened 2023-07-05
Try using workflow from wporg-repo-tools

ryelle:
See https://github.com/WordPress/wporg-repo-tools/issues/26 — this demonstrates updating the actions to use the shared composite actions.

This is still a draft because if https://github.com/WordPress/wporg-repo-tools/pull/29 merges, we'll need to update the action paths to remove the branch name.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,46 +1,39 @@
 name: Build and push to build branch.
 
 on:
-    push:
-        branches: [trunk]
+  push:
+    branches: [trunk]
 
 jobs:
-    build:
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout repository
-              uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: trunk
 
-            - name: Install NodeJS
-              uses: actions/setup-node@5b52f097d36d4b0b2f94ed6de710023fbb8b2236 # v3.1.0
-              with:
-                  node-version-file: '.nvmrc'
-                  cache: 'yarn'
+      - name: Setup
+        uses: WordPress/wporg-repo-tools/.github/actions/setup@try/reusable-workflows
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
 
-            - name: Install all dependencies
-              run: |
-                  composer install
-                  yarn
+      - name: Trim the repo down to just the theme
+        run: |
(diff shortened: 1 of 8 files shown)


ryelle:
I've requested review on this, even though it's a draft, as part of the work on https://github.com/WordPress/wporg-repo-tools/pull/29.

.github/workflows/build.yml
```diff
@@ -1,46 +1,39 @@
...
-                  message: 'Build: ${{ github.sha }}'
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          ref: trunk
+
+      - name: Setup
```
adamwoodnz, line 16: I guess we won't see this workflow run until merged. Noting there are a couple of extra steps in Setup which don't currently run on build; 'Setup PHP with PECL extension' and 'Setup configs'.

adamwoodnz: Oh looks like 'Setup configs' covers the removed `setup:tools` call

adamwoodnz (commented):
Looks like the workflows ran correctly on the branch, except for `build` obviously. Just one comment.

.github/workflows/build.yml
ryelle: > Setup PHP with PECL extension

Previously it was just using the PHP/composer that exist on the container, but "Setup PHP with PECL extension" sets up a specific version & adds the token for composer to use.

> Oh looks like 'Setup configs' covers the removed setup:tools call

Yeah, I guess I probably could switch the command in this step to be `setup:tools`, writing it out explicitly was left over from the original workflow testing. It would add another required script in each project, but I think we're consistently using `setup:tools` (it would also mean I can remove the textdomain input, since each project's setup:tools should handle that).

ryelle: I ended up switching to `setup:tools` in https://github.com/WordPress/wporg-repo-tools/pull/29/commits/31fdc718b776e4b62382a642a35245dfde354e72, and re-ran the checks, but then it was installing yarn & composer twice. I've updated the local `setup:tools` command so that all it does is set up the configs, and that works. It means you need to manually run a few more commands to set up the project, but since we'll only be doing that once per project, it should be okay.

adamwoodnz: Could we do the opposite; remove the `Install all dependencies` step and revert removing the dependency installation from `setup:tools`?

ryelle: I'd rather not because that makes the setup:tools command pretty complicated, since it would need to do the `composer install || composer update wporg/*` step… and I'm not actually sure how well that step would join with the yarn command (like, can you just do `composer install || composer update wporg/* && yarn` or is there an order of operations thing going on).

ryelle:
This is fixing the current build issue in trunk (because the wporg-mu-plugins repo & other dependencies were out of date), so I'm going to merge it.

Issue #266, opened 2023-05-15
Update rosetta homepages to new homepage version


StevenDufresne:
The rosetta site homepages are still showing the older version. They were not updated during the initial rollout of the new homepage to limit the scope. It's time to update the rest.

dd32:
#15 covers the background to why this wasn't initially updated.

ryelle:
We've had the content parser running for a while, and it's worked fairly well (especially with the updates and tests from https://github.com/WordPress/wporg-main-2022/pull/247 & https://github.com/WordPress/wordpress.org/pull/90).

This project is not imported into GlotPress yet, that would need to happen so it could be translated— we should let #polyglots know that this is incoming first, in case there's more feedback on the strings & process.

Then maybe soft-launching with the theme switcher again? So we can get feedback on it in-action? Maybe that's overkill.

I tested wporg-main-2022 on my sandbox to see how it holds up, and mostly it was fine.

My notes from that test (these can become issues if needed, but right now it's just a brain-dump).

- wporg-main-2022 and wporg-parent-2021 will both need to be added to GlotPress
- Do rosetta sites use localized page slugs?
    - Looks like no, but it might affect the template assignment if so.
- For the local pages (those that are specific to the rosetta site, not run through GlotPress), could we just turn on the default redesign page template? (like what the About subpages use)
- Needs to support posts & archives, [example on ja.w.org](https://ja.wordpress.org/news/), [example on es.w.org](https://es.wordpress.org/news/)
    - The sidebar can be unique to a site, by updating the widgets. How important is this? Can we allow some customization without giving access to the entire Site Editor?
- In-page navigation menus (About, Download)
    - Currently hard-coded to a menu ID on wp.org, won't exist on rosetta site.
    - We could either loose the menu and hardcode the items in the template, then set up a filter like what's [run in the global header](https://github.com/WordPress/wporg-mu-plugins/blob/523df8442190613807c098b1ffc60d81a9436898/mu-plugins/blocks/global-header-footer/blocks.php#L880) to swap out the locale URL
    - Use `__unstableLocation`, and have each rosetta site set a menu to each location. but it's "unstable"
    - Add a custom attribute to mirror ^ but since we set it, it will be stable.
- Need to update links to locale specific
	- Some links in content could be translated in the string
		- Would work for “Get started” on home
		- But not for About page list, hrefs are not part of the list item string (maybe they should be after all)
	- Nav on Download page (below Downlod button)
[...]

PR #288, opened 2023-07-20
Navigation Block: Add option to show a hardcoded "dynamic menu"

ryelle:
See #266 — Core navigation menus work by using a `ref`, an ID to a menu post. On the main site, this is fine, but on rosetta sites, this does not work because that post is not available on the local site.

For example, the About subpage should have a right-hand menu:

| English | Rosetta |
|---|---|
| Screen Shot 2023-07-20 at 15 34 51 | Screen Shot 2023-07-20 at 15 34 44 |

We could do some switch_to_blog work, but that would still load the English menu & links to the English site. Instead, I've added a new attribute to the Navigation block, `menuSlug`. When this is present (and not empty), the menu's content is swapped out for a hardcoded menu which uses the local site's base URL and wraps the menu text in translation wrappers.

I also added some styles for the vertical nav menu format, to match the list of links style, so we can use that on the About page (otherwise the links to the main English site were hardcoded).

### Screenshots

Each menu in these screenshots are dynamically generated, linking to the page on the rosetta site.

| Label | Screenshot |
|---|---|
| About | Screen Shot 2023-07-20 at 15 45 45 |
| About subpage | Screen Shot 2023-07-20 at 15 42 59 |
| Download | Screen Shot 2023-07-20 at 15 43 14 |
| Download subpage | Screen Shot 2023-07-20 at 15 43 25 |

The dynamic menu doesn't have an in-editor preview, but the menu slug can be controlled in the editor — this at least prevents any regressions when the content is updated.

nav-in-editor

1. About, Download, and the subpages should have visible menus regardless of whether those IDs exist on your local site.
2. Create or edit a page with a Navigation block on it
3. Switch it to use a dynamic menu
4. You should be able to control all the other settings (ex, layout & styles)
5. It should render that dynamic menu on the front end, with the settings applied


adamwoodnz:
On insert the nav block chooses an existing menu, then when I change to one of the new dynamic options the block preview keeps the initial items. Can we update that?

dynamic nav

adamwoodnz (commented):
LGTM, just one quirk with the Editor view of the dynamic menu

ryelle:
> On insert the nav block chooses an existing menu, then when I change to one of the new dynamic options the block preview keeps the initial items. Can we update that?

I know 😞  I couldn't figure out a way around that, because if we delete the `ref` to the real menu, GB just tries to be smart and "fixes" the missing `ref` ID. If it's set to something arbitrary, the whole block fails to load. On the frontend, it should use the dynamic menu you selected despite the menu displayed in the editor.

I'll make a follow-up issue to figure this out, but I don't think it's a high priority fix since most people won't be editing pages with menus on them (the About & Download header menus have been moved out of the page content, so only the Download & About landing pages have them).

adamwoodnz (approved):
> On the frontend, it should use the dynamic menu you selected despite the menu displayed in the editor.
> 
> I'll make a follow-up issue to figure this out, but I don't think it's a high priority fix since most people won't be editing pages with menus on them (the About & Download header menus have been moved out of the page content, so only the Download & About landing pages have them).

💯 LGTM

PR #289, opened 2023-07-20
Parser: Add a note for translators when a string contains a shortcode 

### LINKED ISSUE #284: When using shortcodes/variables in strings - Add a translator note.
Since the raise of AI lately, a lot of polyglots are using AI translation. Yet, AI translation sometimes translates shortcodes/placeholders like in this string: https://translate.wordpress.org/projects/meta/wordpress-org/nl-be/default/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=16046810&filters%5Btranslation_id%5D=105620034

This might break the website in some locales and would case a mess. Shouldn't we add translator notes to NOT translate these words between [  ] ? 

Here is an example list:
https://translate.wordpress.org/projects/meta/wordpress-org/nl-be/default/?filters%5Bterm%5D=%5B&filters%5Bterm_scope%5D=scope_originals&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=translation_date_modified&sort%5Bhow%5D=desc

ryelle:
Fixes #284 — This updates the content parser to add a translator comment on strings with shortcodes.

https://github.com/WordPress/wporg-main-2022/blob/7a56dec288104f32a0feb3daec0a01105dd56683/source/wp-content/themes/wporg-main-2022/patterns/about-requirements.php#L18-L31

1. Run the phpunit tests
2. Build the patterns with `yarn build:patterns` (should be no change since this was done in https://github.com/WordPress/wporg-main-2022/commit/e962cc3984b7df09a319eaad0106673237261bce, except for the 6-3 page which is still in progress)

### DIFF
--- a/env/export-content/includes/parser.php
+++ b/env/export-content/includes/parser.php
@@ -236,11 +236,25 @@ function replace_with_i18n( string $content, string $textdomain = 'wporg' ) : st
 
 	$i18n_strings = [];
 	foreach ( $strings as $string ) {
-		$i18n_strings[ $string ] = sprintf(
-			"<?php _e( '%s', '%s' ); ?>",
-			str_replace( "'", '&#039;', $string ),
-			$textdomain
-		);
+		if ( preg_match_all( '#\[[a-z_-]{5,}\]#', $string, $matches ) ) {
+			if ( count( $matches[0] ) > 1 ) {
+				$translator_comment = sprintf( '/* translators: %s are shortcodes and should not be translated. */', implode( ', ', $matches[0] ) );
+			} else {
+				$translator_comment = sprintf( '/* translators: %s is a shortcode and should not be translated. */', implode( ', ', $matches[0] ) );
+			}
+			$i18n_strings[ $string ] = sprintf(
+				"<?php\n%s\n_e( '%s', '%s' );\n?>",
+				$translator_comment,
+				str_replace( "'", '&#039;', $string ),
+				$textdomain
+			);
+		} else {
+			$i18n_strings[ $string ] = sprintf(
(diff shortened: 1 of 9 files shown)


adamwoodnz (commented):
LGTM, just some lint errors

dd32 (approved):
LGTM

dd32:
The linting errors seem.. silly IMHO. `Squiz.PHP.EmbeddedPhp.ContentBeforeOpen`, `Squiz.PHP.EmbeddedPhp.ContentAfterEnd`

I'm personally in favour of excluding these where we exclude their siblings here:
https://github.com/WordPress/wporg-repo-tools/blob/trunk/configs/phpcs.xml.dist#L28-L31

Instead of 
```diff
- <p><?php _e( .... ); ?></p>
+ <p><?php
+ /* translators: [market_share] is a shortcode and should not be translated. */
+ _e( .... );
+ ?></p>
```

it wants
```diff
- <p><?php _e( '..... ?></p>
+ <p>
+ <?php
+ /* translators: [market_share] is a shortcode and should not be translated. */
+ _e( .... );
+ ?>
+ </p>
```

Easily fixed with `\n`s but just seems like a weird sniff.

ryelle:
I think I'll exclude that rule for the pattern files ([like this](https://github.com/WordPress/wporg-main-2022/blob/023bc93ccc223dcb34b7f7319d36092d57cd7941/phpcs.xml#L12-L15)), but leave the shared projects config alone for now.

adamwoodnz (approved):
Still LGTM

PR #291, opened 2023-07-26
Parser: Enable translating link destinations

ryelle:
This updates the button parser and list item parser to allow translating link destinations as well as link text.

For buttons, the block parser can interpret the destination as its own item to translate, so only the link is wrapped in `_e()`.

```php
<div class="wp-block-button">
  <a class="wp-block-button__link wp-element-button" href="<?php _e( 'https://wordpress.org/hosting/', 'wporg' ); ?>">
    <?php _e( 'See all recommended hosts', 'wporg' ); ?>
  </a>
</div>
```

For list items, it makes more sense to fall back to the same behavior as paragraphs and other default blocks, where the whole link element is translated.

```php
<li><?php _e( '<a href="https://learn.wordpress.org/course/getting-started-with-wordpress-get-setup/">WordPress courses ↗</a>', 'wporg' ); ?></li>
```

See #266.

1. Check that the tests pass
2. Try re-running `yarn build:patterns`
3. There should be no visual changes to updated pages


env/export-content/includes/parser.php
```diff
@@ -148,7 +148,14 @@ public function to_strings() : array {
...
 
-		return array_unique( $strings );
+		$strings = array_unique( $strings );
+		$strings = array_filter( $strings, array( $this, 'filter_out_shortcodes' ) );
+
+		return $strings;
+	}
+
+	public function filter_out_shortcodes( string $string ) : bool {
+		return ! preg_match( '#^\[[a-z_-]{5,}\]$#', $string );
```
adamwoodnz, line 158: Does the `5` here relate to a minimum length for shortcodes?

adamwoodnz (approved):
LGTM

env/export-content/includes/parser.php
ryelle: Yeah. This came from @dd32's [suggestion for the other shortcode parsing issue](https://github.com/WordPress/wporg-main-2022/issues/284#issuecomment-1635159171), copied over so both shortcode checks work the same way.

PR #307, opened 2023-08-09
Page Meta: Allow Jetpack to grab the featured image if one exists

ryelle:
The 6.3 page has a featured image for use when sharing on social media, but the existing `jetpack_images_pre_get_images` filter short-circuits out of using it. This updates the function to check if there is a thumbnail first.

This function was added to restrict queries on pages without featured images (#62). If you apply this, you should not see queries increase on those pages.

To test:

1. Add a featured image to a page
2. View source, the `og:image` should use the featured image


adamwoodnz (approved):
👍 Works for me

Issue #245, opened 2023-03-30
Create 'Blocks' Landing Page


bengreeley:
Currently, there isn't a page on wordpress.org that explains in a compelling way what 'Blocks' are or markets it very effectively on the website. We have a [filter](https://wordpress.org/plugins/browse/blocks/) in the plugin directory, which is useful, but that page is lacking the context of what blocks are, what the block editor is, and why it is so exciting. There are other pages, such as the [documentation](https://wordpress.org/documentation/article/wordpress-block-editor/), that explains what the block editor is, but IMO it's a bit wordy and explains how to use the editor, but not necessarily _why_ you should use the editor and why it is unique and exciting.

It would be great to think through a landing page that explains the block editor and links to the various resources, the listing of blocks, etc. I envision something more visual like a marketing page for a product vs being text-heavy, but I'll leave the actual implementation ideas to folks who are more creative than I am. :-) 

cc @WordPress/meta-design @thetinyl @StevenDufresne

jasmussen:
Connecting dots with https://github.com/WordPress/wporg-mu-plugins/issues/360, which discusses whether to rename the "Download & Extend" menu item.

fcoveram:
My first thought about this content is addressing it across multiple pages. Landing pages tend to live isolated when they are feature-focused and intend to convey a product value. A landing page definitely works for diving into details and making statements more precise. But up to now, I envision block references in the homepage.

I would add this content request to the IA mapping to then drop ideas with comms folks for updates on multiple pages.

StevenDufresne:
I would consider how this relates to http://wordpress.org/gutenberg. It seems like there would be some overlap and the block aspect of /gutenberg holds that page together since it primary talks about using the block to make pages.

jasmussen:
That consideration would be good to consider in the design process for a blocks page. The Gutenberg page could potentially be linked from the blocks page mainly (or only), or it could even be partially embedded inside. That said, my instinct at this point is that the current separate destination is valuable to still have.

thetinyl:
Just some immediate questions that came up for me thinking about a blocks page:

- What value does a separate Blocks landing page have for folks new to WordPress?
- Who would this landing page be for? 
- If blocks/block editor is modern WordPress, how does a blocks landing page help knit together the story of the WordPress experience as a whole?

My initial thoughts are pretty similar to @panchovm. 

> My first thought about this content is addressing it across multiple pages.

[...]

PR #313, opened 2023-08-12
Theme Switcher: Default to using the redesigned theme for pages

ryelle:
This will reverse the logic for loading the redesign theme, so that pages are opted-out of the new theme, rather than opted in. It also makes the page check case-insensitive. This fixes some edge cases where technically-valid page URLs are used, but the theme switcher fails to load the correct template.

For example, the following URLs should now load the correct theme & content. 

- Capitalized slugs: `wordpress.org/About` (or Blocks, as the original issue found, but there's a workaround for that page in place now)
- Slugs with dots: `wordpress.org/download/releases/6.3/` (IIRC this was reported in slack but not followed up on)
- Arbitrary dashes: `wordpress.org/download/beta----nightly/`
- More dots: `wordpress.org/download/beta....nightly/`

Ideally these URLs should redirect to the real canonical URL, but since they are functional links, the theme should at least not be broken. There are core tickets for [the arbitrary dashes](https://core.trac.wordpress.org/ticket/14773) and [dots in permalinks](https://core.trac.wordpress.org/ticket/35437) (which also applies to the 6.3 issue despite seeming related to the post title).

See https://github.com/WordPress/wporg-main-2022/issues/245#issuecomment-1675270164, https://meta.trac.wordpress.org/ticket/7204

1. Check out the branch, test locally or sync to sandbox (`mu-plugins/main-network/theme-switcher.php`)
2. Try visiting any of the example URLs
3. The page should load the new theme and show the correct content
4. Try other valid pages, they should continue to work
5. Visit `/hosting/` or another "old theme" page
6. It should still use the old theme & content

### DIFF
--- a/source/wp-content/mu-plugins/theme-switcher.php
+++ b/source/wp-content/mu-plugins/theme-switcher.php
@@ -20,27 +20,12 @@ function should_use_new_theme() {
 		return false;
 	}
 
-	// Request to resolve a template.
-	if ( isset( $_GET['_wp-find-template'] ) ) {
-		return true;
-	}
-
-	$request_uri = isset( $_SERVER['REQUEST_URI'] ) ? explode( '?', esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ?? '/' )[0] : '/';
-
-	// Admin page or an API request.
-	if ( is_admin() || wp_is_json_request() || 0 === strpos( $request_uri, '/wp-json/wp' ) ) {
-		return true;
-	}
-
 	// Preview. Can't call is_preview() this early in the process.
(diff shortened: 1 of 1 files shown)


adamwoodnz (commented):
Works for me.

Wondering if we need all of the truthy conditions above anymore, now that more pages will already get the new theme by not being in the exception list, for example `/search/` and maybe `preview` links?

ryelle:
Good idea, that does simplify the function. Looks like we don't need the search one, but I kept preview so that those links will always use the new theme.

adamwoodnz (approved):
:shipit:  Still LGTM

PR #329, opened 2023-10-12
Facebook now capped

### LINKED ISSUE #328: Facebook logo not capped?
https://github.com/WordPress/wporg-main-2022/blob/53a72ba1035dc2d69b9ab9d7225cb08d5c45231a/source/wp-content/themes/wporg-main-2022/patterns/enterprise.php#L67

On this page, all company names are capped... Apart from Facebook. 
I think this is just somebody being a goof and being distracted... 

(In Dutch we have a nice word for that. It's "verstrooid" and it basically means to do something be accident or when you make a mistake when you are absent-minded. "Ik was verstrooid" means "I was absent-minded")

NekoJonez:
While checking the new translations for WP.org, I noticed that Facebook logo wasn't capped like all the other companies. So, I fixed that in this PR.

Would fix #328


NekoJonez:
Fixed via #330

NekoJonez:
Gonna close this one, since it's fixed via another.

PR #346, opened 2023-11-08
Add a remembers block.

StevenDufresne:
This PR implements a block that will query wp.org for memorialized profiles.

## Editor 

Screenshot 2023-11-08 at 3 11 48 PM

## Front End
localhost_8888_remembers_ (1)


source/wp-content/themes/wporg-main-2022/src/remembers-list/index.php
```diff
@@ -57,7 +62,7 @@ function render( $attributes, $content, $block ) {
 			$block_content .= '<!-- wp:heading {"textAlign":"center","style":{"spacing":{"margin":{"top":"var:preset|spacing|30","right":"var:preset|spacing|default","bottom":"var:preset|spacing|30","left":"var:preset|spacing|default"}}},"fontSize":"extra-large"} -->';
 			$block_content .= '<h2 class="wp-block-heading has-text-align-center has-extra-large-font-size" style="margin-top:var(--wp--preset--spacing--30);margin-right:var(--wp--preset--spacing--default);margin-bottom:var(--wp--preset--spacing--30);margin-left:var(--wp--preset--spacing--default)">';
 			$block_content .= '<em>';
-			$block_content .= '<a href="'. esc_url( $profile["link"] ) .'">' . esc_html( $profile["name"] ) . '</a>';
+			$block_content .= '<a href="https://profiles.wordpress.org/'. esc_attr( $profile->user_nicename ) .'">' . esc_html( $profile->display_name ) . '</a>';
```
pkevan: we'd want to run `esc_url` on the full URL to escape it, `esc_attr` is for full html attribute values, so:

```
esc_url( "https://profiles.wordpress.org/" . $profile->user_nicename )
```

StevenDufresne: Yep. Thanks

source/wp-content/themes/wporg-main-2022/src/remembers-list/index.php
```diff
@@ -0,0 +1,80 @@
...
+	$group_count = ceil( count( $profiles ) / $columns );
+
+	$groups = [];
+	for ( $i = 0; $i < $group_count; $i++ ) {
+		$groups[] = array_slice( $profiles, $i * $columns, $columns );
+	}
+
+	$block_content = '';
+	foreach ( $groups as $group ) {
+		// Set isStackedOnMobile to false so that the columns are not stacked on mobile. We override this in CSS to stack them.
```
adamwoodnz, line 56: Expand to include why we override? Is it because we want a specific breakpoint?

adamwoodnz: I haven't tried but is it possible to override a custom property for the breakpoint scoped to this block, so that we can use the built in responsive behaviour?

adamwoodnz: It gets a bit tight around 540px wide, might need some edge spacing

StevenDufresne: > Expand to include why we override? Is it because we want a specific breakpoint?

It's either not use the column block at all and control everything using css, otherwise, yeah it breaks way too early.

StevenDufresne: I don't see that possibility. 
- https://github.com/WordPress/gutenberg/issues/35848
- https://github.com/WordPress/gutenberg/issues/56138
- https://github.com/WordPress/gutenberg/blob/26d96b6fbc8ff13544c0042ffc209d1aead7b20f/packages/block-library/src/columns/style.scss#L8
- https://github.com/WordPress/gutenberg/blob/26d96b6fbc8ff13544c0042ffc209d1aead7b20f/packages/base-styles/_breakpoints.scss#L11

Maybe I'm looking in the wrong spot.

adamwoodnz: All good, thanks for looking

StevenDufresne: I'm going to merge as is. Thanks!

PR #356, opened 2023-11-28
Add trailing slash to `remembers-list` profile URLs

ironprogrammer:
The profile page links emitted by the `remembers-list` block don't have a trailing slash, yielding e.g. `https://profiles.wordpress.org/viper007bond`.

In order to mitigate a redirect when accessing the profile, this PR adds a trailing slash, e.g. `https://profiles.wordpress.org/viper007bond/`. The URL is composed with `sprintf` for readability to see where the profile name substitution occurs.

1. Load the `/remembers/` page.
2. Observe the profile link HREFs for each name. They should each have a trailing slash after the username.


pkevan (approved):
Thanks @ironprogrammer - looks good.

PR #366, opened 2023-12-13
Update home.php

pkevan:
Switch location to events homepage.


adamwoodnz:
I've just updated the content in the editor to match. It's important to do this, otherwise the next time we generate the pattern the change could be overwritten 🙂

pkevan:
> I've just updated the content in the editor to match. It's important to do this, otherwise the next time we generate the pattern the change could be overwritten 🙂

Thanks - i'd assumed that updating the template would cause it to change, but noted for next time.

PR #374, opened 2023-12-20
Data Liberation: Add a data liberation handbook setup

dd32:
This PR adds the necessary things to the theme to support the import of the data liberation guides as a handbook.

See https://github.com/WordPress/data-liberation/pull/46 for the manifest it uses.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -15,7 +15,9 @@
 	"plugins": [
 		"./source/wp-content/plugins/gutenberg",
 		"./source/wp-content/plugins/jetpack",
-		"./source/wp-content/plugins/wordpress-importer"
+		"./source/wp-content/plugins/wordpress-importer",
+		"./source/wp-content/plugins/handbook",
+		"./source/wp-content/plugins/wporg-markdown"
 	],
 	"themes": [
 		"./source/wp-content/themes/wporg",
--- a/composer.json
+++ b/composer.json
@@ -80,6 +80,26 @@
 						"url": "https://meta.svn.wordpress.org/sites/",
 						"reference": "trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/"
 					}
+				},
+				{
+					"name": "wordpress-meta/handbook",
+					"type": "wordpress-plugin",
+					"version": "1",
+					"source": {
+						"type": "svn",
+						"url": "https://meta.svn.wordpress.org/sites/",
+						"reference": "trunk/wordpress.org/public_html/wp-content/plugins/handbook/"
+					}
+				},
+				{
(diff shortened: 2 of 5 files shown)


dd32:
I don't think cron tasks work correctly in the local environment, so you'll probably have to run them manually when required:

```
yarn wp-env run cli wp cron event run --due-now
```

dd32:
Just noting, if we want to commit/deploy this prior to resolving all style issues, and as a result have the `/and/{guide}` route not work for non-editors, we can disable it from loading with something like this:
```
add_filter( 'handbook_post_type_defaults', function( $args, $slug ) {
    if (
	    'and' === $slug &&
	    ! current_user_can( 'edit_handbook_pages' ) &&
	    ! wp_doing_cron() &&
	    ! ( defined( 'WP_CLI' ) && WP_CLI )
    ) {
	    return false;
    }
    
    return $args;
}, 10, 2 );
```

Additionally, we probably also want to add a check in here that only loads it for not-rosetta sites, as we currently have no way to translate handbooks (which is a separate issue, and this is not going to resolve)

_edit: in retrospect, just not enabling the handbook plugin on rosetta will result in that, so nevermind!_

StevenDufresne:
The only question I have is the url structure. 

We currently have `wp.org/data-liberation` as the landing page. 

I imagine we could change `/data-liberation` to `/and` so then we could have:
- `wp.org/and`
- `wp.org/and/{platform}`

I don't know that the following reads that well:
- `wp.org/data-liberation` 
- `wp.org/data-liberation/and` 
- `wp.org/data-liberation/and/{platform}` 

However we do need the following:
- Landing page
- Archive page
- Single page

So maybe it makes sense to make `/data-liberation` the root.

dd32:
Yeah I wasn't sure on that either :)

The good news is that we can change the base slug easily.

Currently the PR for the handbook simply sets `w.org/and/` as a post type index, but as you've probably seen, creating a specific page in GitHub (`welcome`, `index`, `handbook`, or `and-handbook`) would place that at `w.org/and` instead. We probably don't want to use a specific page like that though, we probably just want the post-type archives and our custom landing pages.

Based on the WordPress/data-liberation contents, it might make sense to split it up into a few different "sections" of content, based on the intention of the underlying guide:
 - Move to WordPress (from another platform)
 - Move from non-core to Core Blocks
 - Other Tools 

so something like:
 - `w.org/data-liberation/from/` (archive)
 - `w.org/data-liberation/from/{platform}` (single)
 - `w.org/data-liberation/migration-guide/` (archive of the subsections:)
 - `w.org/data-liberation/migration-guide/{plugin}-to-{blocks}` (single)

PR #375, opened 2023-12-20
Data Liberation: Add interior pages

StevenDufresne:
This PR adds an interface for the guide location in the https://github.com/WordPress/data-liberation repository.

 
### Screenshots

| Header | Header |
|--------|--------|
| Home |  |
| Archive | localhost_8888_data-liberation_and_ (1) |
| Single | localhost_8888_data-liberation_and_rss-4_ | 

- run `yarn wp-env run cli wp cron event run --due-now`
  - See this PR for background: https://github.com/WordPress/wporg-main-2022/pull/374 
- Create a page `data-liberation`
  - Associate the `page-data-liberation` template
- Create a child page of `data-liberation` with the slug `and`


source/wp-content/themes/wporg-main-2022/templates/single-and-handbook.html
```diff
@@ -0,0 +1,10 @@
+<!-- wp:wporg/global-header {"style":{"border":{"bottom":{"color":"var:preset|color|white-opacity-15","style":"solid","width":"1px"}}}} /-->
+
+<!-- wp:pattern {"slug":"wporg-main-2022/nav-data-liberation-single"} /-->
+
+
```
outdoor2kode, line 5: extra line.

source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -138,6 +144,25 @@ function override_lazy_loading( $value, $image ) {
...
+				'url' => home_url( '/data-liberation' ),
+				'title' => __( 'Home', 'wporg' ),
+			),
+			array(
+				'url' => home_url( '/data-liberation/and' ),
+				'title' => __( 'Guides', 'wporg' ),
+			),
+			array(
+				'url' => false,
+				'title' => get_the_title(),
```
outdoor2kode, line 162: This would somehow get me the title plus an edit button, is this also the case on your side?

image

source/wp-content/themes/wporg-main-2022/patterns/data-liberation.php
```diff
@@ -30,56 +30,56 @@
 
 <!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|10"}},"className":"is-style-cards-grid","layout":{"type":"grid","minimumColumnWidth":"32%"}} -->
 <div class="wp-block-group is-style-cards-grid"><!-- wp:wporg/link-wrapper -->
-<a class="wp-block-wporg-link-wrapper" href="https://github.com/WordPress/move-to-wp/blob/trunk/guides/squarespace-to-wordpress.md"><!-- wp:heading {"level":3,"style":{"spacing":{"margin":{"top":"0","bottom":"0"}},"typography":{"fontStyle":"normal","fontWeight":"700"}},"fontSize":"small","fontFamily":"inter"} -->
+<a class="wp-block-wporg-link-wrapper" href="<?php echo esc_url( site_url( 'data-liberation/and/squarespace' ) ); ?>"><!-- wp:heading {"level":3,"style":{"spacing":{"margin":{"top":"0","bottom":"0"}},"typography":{"fontStyle":"normal","fontWeight":"700"}},"fontSize":"small","fontFamily":"inter"} -->
```
outdoor2kode: Using `data-liberation/and/squarespace` would get an extra 301 Redirect.
We should add a `/` to the URL =>`data-liberation/and/squarespace/`.

outdoor2kode (commented):
Looks good to me, mostly works as described. 👍  
Left some comments.
And we can consider aligning the post title with the top of the TOC.

image
image

source/wp-content/themes/wporg-main-2022/patterns/data-liberation.php
outdoor2kode: Same for the URLs below.

source/wp-content/themes/wporg-main-2022/functions.php
ryelle: I also see this, looks like it comes from `filter_the_title_edit_link` in the wporg-markdown plugin.

source/wp-content/themes/wporg-main-2022/patterns/data-liberation.php
ryelle: This page is synced from the production site, so the PHP here would be overwritten when the script is run. Should it be removed from the syncing?

source/wp-content/themes/wporg-main-2022/templates/single-and-handbook.html
```diff
@@ -0,0 +1,10 @@
...
+<!-- wp:group {"tagName":"main","style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space","bottom":"var:preset|spacing|70"}}},"className":"alignfull","layout":{"type":"constrained"}} -->
+<main class="wp-block-group alignfull" style="padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--edge-space)">
+<!-- wp:pattern {"slug":"wporg-main-2022/data-liberation-guide-single"} /-->
+</main>
+<!-- /wp:group -->
```
ryelle, line 10:
```suggestion
<!-- /wp:group -->

<!-- wp:wporg/global-footer /-->
```

The single pages are missing the footer.

source/wp-content/themes/wporg-main-2022/templates/archive-and-handbook.html
```diff
@@ -0,0 +1,11 @@
+<!-- wp:wporg/global-header {"style":{"border":{"bottom":{"color":"var:preset|color|white-opacity-15","style":"solid","width":"1px"}}}} /-->
+
+<!-- wp:pattern {"slug":"wporg-main-2022/nav-data-liberation"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0px"}},"className":"entry-content","layout":{"inherit":true,"type":"constrained"}} -->
+<main class="wp-block-group entry-content">
+	<!-- wp:pattern {"slug":"wporg-main-2022/data-liberation-guides"} /-->
```
ryelle, line 7: Are there plans to support translating the data lib pages? If not, we probably don't need to use the pattern process here, since these pages wouldn't be rolled out to rosetta sites. (IIRC we don't have a process for translating handbooks yet, which is why i assume no)

Not a strong suggestion, but it would be one less abstraction, to put the pattern content (query loop, etc) here.

source/wp-content/themes/wporg-main-2022/functions.php
```diff
@@ -27,6 +27,12 @@
 add_filter( 'render_block_data', __NAMESPACE__ . '\update_header_template_part_class' );
 add_filter( 'wporg_block_navigation_menus', __NAMESPACE__ . '\add_site_navigation_menus' );
 
+// Remove table of contents.
+add_filter( 'wporg_handbook_toc_should_add_toc', '__return_false' );
+
+// Remove the edit link from handbook titles.
+add_filter( 'wporg_markdown_should_filter_title', '__return_false' );
```
ryelle, line 35: I wonder why this filter seems to not exist anymore… I would have thought this would prevent the issue @renintw [and I are seeing here](https://github.com/WordPress/wporg-main-2022/pull/375/files#r1463459984). I can't find the filter in the wporg-markdown plugin.

adamwoodnz:
> aligning the post title with the top of the TOC

Can be achieved by setting the top margin of the sidebar container via theme.json, see https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/theme.json#L162

source/wp-content/themes/wporg-main-2022/functions.php
StevenDufresne: Ah, dang. That was a filter I implemented in my local environment in the markdown plugin. Essentially returning early [here](https://github.com/WordPress/wordpress.org/blob/cdf89d38440d54b15b07bcda134fa752ce045cd4/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-editor.php#L78). 

I discussed this with @dd32 and since we don't use the title anywhere anymore, we agreed we could remove the function altogether and I forgot to update.

source/wp-content/themes/wporg-main-2022/templates/archive-and-handbook.html
StevenDufresne: I default to yes. We should support it but understand that rosetta experiences vary right now and these pages may never be viewed in other locales... but still feel it should be localized.

source/wp-content/themes/wporg-main-2022/patterns/data-liberation.php
StevenDufresne: The hardcoded URL is a temporary solution. 

An ideal solution is to implement tags/categories to control that list. 

There are 2 reasons why I resorted to hardcoded items:
- We don't have categories set up
- The `<LinkWrapper>` component that was introduced in https://github.com/WordPress/wporg-main-2022/commit/685b3c8859a9baec7ba006a5bc8c6b0ee3f251bf, doesn't play nicely with the `<PostTemplate>` component.

So no, they should not be removed.

source/wp-content/themes/wporg-main-2022/functions.php
StevenDufresne: Is everyone okay with removing that function from the wporg-markdown plugin?

Context: https://github.com/WordPress/wporg-main-2022/pull/375#discussion_r1464322243

outdoor2kode: If it's not going to be used anywhere, I also think we can remove it.

source/wp-content/themes/wporg-main-2022/patterns/data-liberation.php
ryelle: I don't mean to remove the items themselves, but that the PHP here will be overwritten when `yarn build:patterns` is run again. So it should either be changed to `href="/data-liberation/and/squarespace/"` in the page content & re-synced, or this page removed from `page-manifest.json` and disconnect it from the syncing process.

source/wp-content/themes/wporg-main-2022/templates/archive-and-handbook.html
ryelle: okay, if we assume that it will be available eventually then 👍🏻   — my aim was to not make more work for the translators.

StevenDufresne: That's a good thought and something I haven't considered when developing. Thanks.

PR #392, opened 2024-02-22
When creating translated patterns/templates, use esc_attr_e() for attributes

dd32:
Replaces a TODO in the code with replacing attributes with an escaped variant.

This isn't the cleanest way to achieve the result, but it is the smallest possible code change.

It might look like `href` should have `esc_url( __( .. ) )` but that complicates it with minimal benefit.

Draft as I don't have time to follow up on it completely right now. 
Feel free to take over.

1. `yarn build:patterns`
2. Test.


dd32:
See #628

PR #393, opened 2024-02-22
Don't translate relative links

dd32:
The parser already excludes anchor links, such as `#main` but includes root-relative and page-relative urls.

These likely don't need to be translated, and probably shouldn't be.
These probably shouldn't usually exist in the source-content, but have benefits, such as being untranslated after this change.

Example changes included.

1. `yarn build:patterns`
2. Test.

### DIFF
--- a/env/export-content/includes/parsers/HTMLParser.php
+++ b/env/export-content/includes/parsers/HTMLParser.php
@@ -42,12 +42,31 @@ class HTMLParser implements BlockParser {
 				$found_strings = $matches['string'];
 			}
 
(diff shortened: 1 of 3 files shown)


ryelle (approved):
This update also looks good, and makes sense to me  👍🏻 

I'm assuming the test failure is related to the "the upstream mariadb+wp-env issue" from the other parser PR?

dd32:
> I'm assuming the test failure is related

100%, that issue is now resolved, so I've re-run the unit tests and they're now passing.

PR #396, opened 2024-02-29
Requirements page: Bump recommended MySQL version to 8.0

swissspidy:
See https://core.trac.wordpress.org/ticket/59701#comment:5 where the recommended version in the WordPress readme was updated from 5.7 to 8.0, because version 5.7 reached EOL.

I don't know if this is the right place to make this change for the https://wordpress.org/about/requirements/ page though 🤷 

Props @SergeyBiryukov


szepeviktor:
> version 5.7 reached EOL

https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-44.html

pkevan (approved):

PR #412, opened 2024-04-02
Update about-roadmap.php after 6.5 was released

audrasjb:
See https://meta.trac.wordpress.org/ticket/7548


ryelle:
The changes to these templates are [synced down from the page content](https://github.com/WordPress/wporg-main-2022/blob/trunk/readme.md#updating-an-existing-page), so this should be updated in the editor (otherwise it will be overwritten on the next sync). Do you have access to make the change on wp.org?

audrasjb:
Thanks @ryelle but I only have access to WordPress.org/news, no w.org itself :)

PR #419, opened 2024-04-04
ending space deleted

Presskopp:
Translatable strings should not have a leading or ending space. I assume this one can be deleted safely. If not, please put it outside the string.


ryelle (approved):
This should be fine to remove, it was probably added by mistake in the editor. Thanks!

PR #463, opened 2024-07-04
Content: Add a pattern that can be used for "pages in progress"

ryelle:
Some pages shouldn't be fully public (& indexed by search engines) until they're "finished", but we need them published to sync the content to version control + glotpress. Once the page template and pattern are created though, we can swap out the pattern to a placeholder version. This should keep the editor preview working correctly, while not showing the content to regular viewers.

This approach adds a new pattern `wporg-main-2022/work-in-progress`, which can be swapped out on the page template (I resisted the urge to find an under construction gif 😂).

In this case, the content on the pattern references the 6.6 release page, just in case users land on this page from a search engine. Once the 6.6 page is public, we can change out this text with something more generic (cc @thetinyl). Most people will not see this content though.

The content here is not translated, since these strings are only temporary. If we decide to add a more generic message for the future, we can i18n-ize those.

### Screenshots

@WordPress/meta-design — again most people won't see this, but let me know if you want to change anything.

1. View the /download/releases/6-6 page
2. You should not see the real page content
3. You should see a "coming soon" message

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/patterns/_work-in-progress.php
+++ b/source/wp-content/themes/wporg-main-2022/patterns/_work-in-progress.php
@@ -1,0 +1,28 @@
+<?php
+// phpcs:disable WordPress.Files.FileName -- Allow underscore for pattern partial.
+/**
+ * Title: Page in progress
+ * Slug: wporg-main-2022/work-in-progress
+ * Inserter: no
+ *
+ * Use this pattern to prevent publicly showing pages that are not ready yet.
+ */
+
+?>
+<!-- wp:group {"metadata":{"name":"Page"},"align":"full","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull">
+	<!-- wp:cover {"overlayColor":"light-grey-2","isUserOverlayColor":true,"isDark":false,"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}},"layout":{"type":"constrained"}} -->
+	<div class="wp-block-cover alignfull is-light" style="padding-top:var(--wp--preset--spacing--70);padding-bottom:var(--wp--preset--spacing--70)">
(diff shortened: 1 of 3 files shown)


jasmussen:
Looks good, let's get this out there. If we wanted to really polish it, I'd probably go all-white, with a white footer, and a gray border between header/footer/content. But what's here can work just fine, color-wise, especially if that's easier for you. Text-wise, I'd be tempted to simplify this a bit too. Two quick options, not blockers:

Coming Soon alt

Coming Soon

thetinyl:
I think this page should always be generic (if it needs to exist)—even before the release ships. It's one less thing to remember to update and even if its purpose is to never be found, we might as well make it just a little more useful to the small number of visitors it might get. 

My suggestion for copy:

> **You’re looking for what’s new in WordPress**
>
> Go right to the source:
> - [WordPress News](https://wordpress.org/news/) 
> - [Developer Blog](https://developer.wordpress.org/news/)
> - [Upcoming WordPress events](https://events.wordpress.org/upcoming-events/)


If we wanted to be a little cheeky, we could replace "Go right to the source:" with: 

> **You’re looking for what’s new in WordPress**
>
> Lurk where the updates are:
>
> [insert same links]

We can use other links. I'm suggesting the proposed because News and Dev Blog will include insights into ongoing releases. Events is just another opportunity to drive people to engage with the community (without necessarily asking for contribution).

ryelle:
> If we wanted to really polish it, I'd probably go all-white, with a white footer, and a gray border between header/footer/content.

We can't change the header/footer— or if we do, it will impact the preview of the in-progress page, since the template wrapper needs to stay the same. So this in-progress page may appear with different header-footer combos, for example if we used it on 6.3, it would have a black header & footer.

I've gone ahead with the first copy option, @thetinyl, just to get something deployed soon, but since it didn't sound final I've omitted translation tags. Once you know what the copy should be, I can update it properly.

thetinyl:
>  Once you know what the copy should be, I can update it properly.

My only suggested tweak would be to move the WP News link to the top of the list. Otherwise, I think what's there is good for now. 

Thanks @ryelle!

ryelle:
huh, I have no idea how they got reorganized, I copied the list from your comment. I'll make that update & get them translatable. Thanks!

PR #467, opened 2024-07-08
Add details to the automated content update PR

ryelle:
I was aiming to add screenshots & other context to the content update PRs (e.g., https://github.com/WordPress/wporg-main-2022/pull/464), as I've found myself needing to reply with comments describing the changes and/or linking off to relevant issues (when these exist, this is easier 🙂)

This PR is adds a changed-page list & link to screenshots to the automatically-created PR. The screenshots are added as a link to the action's artifacts file, which is a zip file with full-screen screenshots of the changed pages. Changed pages are detected by updates to the pattern files, which are looked up using the `manifest.json` to get the page slug. The slug is then used to fetch page data from the wp.org REST API, which is how we get the title & permalink.

If there are multiple pages changed, they should all appear in the list, and all screenshots will be in the download.

(Note that this won't work with the 6.6 release page, since the pattern is currently set to the "in progress" pattern #463, it will only show the placeholder pattern.)

### Screenshots

Screenshot 2024-07-08 at 5 41 02 PM

https://github.com/ryelle/wporg-main-2022/pull/1

I tested it on my own fork, changing [the action](https://github.com/WordPress/wporg-main-2022/blob/f62ba5bc326ed62faf5f5dc67b5f0f869a0b5be1/.github/workflows/content-update.yml#L14) to `if: github.repository_owner == 'ryelle'`, & making a change to one of the pattern files. That way, when the automated action runs, it sees the "change" and creates a PR.

You can also test the screenshot command by running `yarn screenshot-changes source/wp-content/themes/wporg-main-2022/patterns/about.php` (or any other pattern files). Make sure to run `yarn` first to get the new puppeteer dependency.

### DIFF
--- a/.github/workflows/content-update.yml
+++ b/.github/workflows/content-update.yml
@@ -25,13 +25,19 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Run the content update
+        id: build-patterns
         run: |
           chmod -R ugo+w source/wp-content/themes/wporg-main-2022/
           yarn wp-env start
+          yarn setup:wp
           yarn build:patterns
-          git diff --name-only source/wp-content/themes/wporg-main-2022/**/*.php | xargs yarn screenshot-changes
+          mkdir "$GITHUB_WORKSPACE/artifacts/"
+          git diff --name-only source/wp-content/themes/wporg-main-2022/**/*.php
+          changelist=$(git diff --name-only source/wp-content/themes/wporg-main-2022/**/*.php | xargs yarn --silent screenshot-changes)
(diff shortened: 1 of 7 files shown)


dd32 (approved):
This makes sense to me!

It's a shame we can't inline the images, but upon checking, GitHub does not support automated image uploads to PRs. This is due to potential for abuse.
The only way I could see that we could achieve that would be to commit them to a specific git branch (which has no code history, just the images) and inline those instead.

ryelle:
>  The only way I could see that we could achieve that would be to commit them to a specific git branch (which has no code history, just the images) and inline those instead.

Yeah, that could work - maybe a future improvement :)

PR #481, opened 2024-07-18
Increase the default minimum PHP version shortcode to "7.2"

sirreal:
WordPress 6.6 changed the minimum supported PHP version to PHP 7.2.
[The requirements page continues to claim 7.0+:](https://wordpress.org/about/requirements/)
WordPress also works with PHP 7.0+ and MySQL 5.5.5+.
[The compatibility page notes that WordPress 6.6 dropped support for versions 7.0 and 7.1.](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/)

Fixes https://meta.trac.wordpress.org/ticket/7723

Props pkevan

### DIFF
--- a/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
+++ b/source/wp-content/themes/wporg-main-2022/inc/shortcodes.php
@@ -37,7 +37,7 @@ add_shortcode(
 add_shortcode(
 	'minimum_php',
 	function() {
-		return defined( 'MINIMUM_PHP' ) ? MINIMUM_PHP : '7.0';
+		return defined( 'MINIMUM_PHP' ) ? MINIMUM_PHP : '7.2';
 	}
 );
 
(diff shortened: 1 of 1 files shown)


sirreal:
I've tried running this locally, but I don't see the shortcode value updated 😕 

It looks like this change should be sufficient, I'm not sure what the issue is:

https://github.com/WordPress/wporg-main-2022/blob/ac37e0b9df10b3f3fcc9f2833359b050e1bedf12/source/wp-content/themes/wporg-main-2022/patterns/about-requirements.php#L46-L51

Clorith:
The change noted here will modify the default value if the `MINIMUM_PHP` constant is not defined, if not then that constant needs to be modified, this is not declared in the theme, but rather in the Serverhappy config, found at https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/servehappy-config.php (this means that any patch for this particular file will need ot be written for trac, and not GitHub :) )

sirreal:
Thanks for linking to that! I've attached a patch to the ticket.

javiercasares:
I don't know if it can be possible, but it's not really 7.2, but 7.2.24 or 7.2.24 (depends on the source) the minimum version…

https://make.wordpress.org/core/2024/04/08/dropping-support-for-php-7-1/
https://make.wordpress.org/hosting/2024/07/10/wordpress-6-6-server-compatibility/

PR #537, opened 2025-02-04
updates some links

timse201:
moves to https and removes some documentation redirects


ryelle:
Thanks for the PR. Most of the changes look fine, but can you remove the changes in the `patterns/` directory? Those files are automatically generated from the production site's content, so it needs to be edited on WordPress.org. If you want to propose changes to any content there, please open a separate issue.

If you have the dev environment running locally, you should be able to run `yarn build:patterns` to re-sync the pattern content.

timse201:
it's now reverted
my coding skills are rudimentary and are enough to understand the code as a polyglot ;)

this one?
https://github.com/WordPress/wordpress.org/tree/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-main
seems like page-download.php is outdated

PR #544, opened 2025-02-17
Update package.json devDependency versions

dd32:
While investigating #541 and #542 I tried updating the modules, I guess we should probably do that at some point? I don't know what (if anything) this would break/fix.

### DIFF
--- a/.github/workflows/content-update.yml
+++ b/.github/workflows/content-update.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
         with:
-          ref: try/update-modules
+          ref: trunk
 
       - name: Setup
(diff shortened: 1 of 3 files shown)


ryelle:
> I guess we should probably do that at some point? I don't know what (if anything) this would break/fix.

A quick glance at the changelogs, not much for env, some new linter rules in eslint-plugin, scripts updated stylelint, and I don't know about puppeteer — I think we're just using that for the screenshot changes though.

The lint error here is due to the scripts' stylelint upgrade, I have a PR into the repo-tools repo to update the stylelint rules: https://github.com/WordPress/wporg-repo-tools/pull/32

Is it possible to tell composer to install that branch of repo-tools? I don't think we can merge that until all the projects can be updated. Actually… if we switch to a js config file for stylelint, maybe we can check the scripts version and conditionally load the correct rules… I can try that out in repo-tools this week.

dd32:
Replaced with #627

PR #602, opened 2025-12-06
Recommend MariaDB due to MySQL 8.4+ incompatibility

ottok:
MySQL 8.4 and newer are incompatible with WordPress because they have removed support for `SQL_CALC_FOUND_ROWS`. Thus the server requirement recommendations needs to be updated to prioritize MariaDB.

To prevent users from setting up sites on incompatible database versions, this change updates the content to:

- Recommend MariaDB as the primary database solution.
- Specify that MySQL versions 8.0 or older are compatible where MySQL is mentioned.
- Update general references from "PHP and MySQL" to "PHP and MariaDB" or "PHP and MariaDB/MySQL" for accuracy.

This change addresses the compatibility issue and reflects MariaDB's widespread adoption within the hosting ecosystem and its alignment with the open source ethos of the WordPress project. Continued promotion of Oracle's MySQL and indirectly Oracle's expanding media ecosystem is not in the interest of WordPress and the ethos of "democratize publishing".

This affects the contents of the pages https://wordpress.org/about/requirements/ and https://wordpress.org/download/.

As seen on https://wordpress.org/about/stats/#mysql_version, over 60% of WordPress installations today use MariaDB:

image

### DIFF
--- a/env/data.xml
+++ b/env/data.xml
@@ -719,7 +719,7 @@
 <!-- /wp:heading -->
 
 <!-- wp:paragraph -->
-<p>We recommend servers running version [recommended_php] or greater of <a href="https://www.php.net/">PHP</a> and <a href="https://www.mysql.com/">MySQL</a> version [recommended_mysql] <em>OR</em> <a href="https://mariadb.org/">MariaDB</a> version [recommended_mariadb] or greater.<br>We also recommend either <a href="https://httpd.apache.org/">Apache</a> or <a href="https://nginx.org/">Nginx</a> as the most robust options for running WordPress, but neither is required.</p>
+<p>We recommend servers running version [recommended_php] or greater of <a href="https://www.php.net/">PHP</a> and <a href="https://mariadb.org/">MariaDB</a> version [recommended_mariadb] or greater.<br>We also recommend either <a href="https://httpd.apache.org/">Apache</a> or <a href="https://nginx.org/">Nginx</a> as the most robust options for running WordPress, but neither is required.</p>
 <!-- /wp:paragraph --></div>
 <!-- /wp:column -->
 
--- a/env/export-content/tests/block-parser-test.php
+++ b/env/export-content/tests/block-parser-test.php
@@ -170,8 +170,8 @@ class BlockParser_Test extends WP_UnitTestCase {
 	public function data_block_content_i18n_with_shortcode() {
 		return [
 			[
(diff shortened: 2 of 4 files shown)


dd32:
> MySQL 8.4 and newer are incompatible with WordPress because they have removed support for SQL_CALC_FOUND_ROWS.

Do you have a reference for this?

It looks like MySQL 8.4 has **deprecated** the functionality, and not removed it.

> The SQL_CALC_FOUND_ROWS query modifier and accompanying [FOUND_ROWS()](https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_found-rows) function are deprecated; expect them to be removed in a future version of MySQL
> _https://dev.mysql.com/doc/refman/8.4/en/information-functions.html#function_found-rows_

I would suggest this be raised in Core first before this documentation update suggestion is made, and a discussion occur on whether Core should remove `SQL_CALC_FOUND_ROWS` in favour of the double-query, both to support MySQL ?.?+ when it's removed, and to better improve support for other DB systems (For example, SQLite doesn't support this type of query, and you have to do the MySQL recommended double-query)

ottok:
In the link above to WordPress stats you can see that MySQL 8.4 share is only 0.9% and all newer versions than that are 0%. This PR is not a suggestion for "Core should remove SQL_CALC_FOUND_ROWS", why are you interpreting it like that?

This is a suggestion to simply update the docs to have the requirements page and recommendations list MariaDB as the primary and fully compatible option.

dd32:
> This PR is not a suggestion for "Core should remove SQL_CALC_FOUND_ROWS", why are you interpreting it like that?

That's my suggestion in response to your suggestion.

WordPress fully supports MySQL, and MariaDB, at this time.
It's more likely that Core will update to support MySQL changes, than not to.
Suggesting that WordPress requirements are MariaDB over MySQL here will cause no benefit to end-users, and only cause confusion. While yes, I note it's a recommendation and not a requirement paragraph, I still feel that it just adds confusion.

ottok:
Why would it add confusion? Automattic is a MariaDB Foundation sponsor, so one would think that recommending MariaDB makes sense, and as shown in the stats MariaDB is already 60%+ of the install base? Promoting Oracle's products and indirectly their expanding media empire seems going against WordPress' motto to _democratize publishing_.

Can you please reopen this to allow further opinions/views/discussion?

PR #606, opened 2025-12-17
Improve local environment setup reliability

t-hamano:
I ran `yarn setup:wp` to set up my local environment and encountered trhee problems. This PR fixes them.

### Increase the timeout period when retrieving remote content

The default timeout for `wp_remote_get` is `5`, but in my environment (WSL2), an error occurs because the response cannot be received within that time. In my environment, the average time required to retrieve the content was around 7 seconds. Perhaps 5 seconds may be too short in other environments, so it would be good to extend this.

### Don't use hard-coded IDs when configuring front page

I noticed that on the production site, the ID of the page with the slug `home` is no longer `8891`. Instead of using a hard-coded ID, find the page with the slug `home` and set it as the front page.

### Set front page correctly on first run

The `yarn setup:wp` scripts are executed in the order `setup.sh`, `refresh.sh`. `setup.sh` attempts to set the front page, while `refresh.sh` imports content. This means that if there are no pages on the local site yet, setting the front page will fail.

For this reason, I separated the `setup.sh` and `refresh.sh` scripts and add code to the `setup.sh` script to import content.

- Change the homepage setting to "Your latest posts".
- Delete all the pages.
- Run `yarn setup:wp`.
- The script will finish successfully.
- Pages will be generated correctly.
- The page with the `home` slug will be set as the front page.


env/import-content.php
```diff
@@ -73,6 +73,7 @@ function import_rest_to_posts( $rest_url ) {
 			'timeout' => 60,
 		)
 	);
+
```
dd32:
```suggestion
```

env/setup.sh
```diff
@@ -13,4 +13,14 @@ wp option update blogdescription "Blog Tool, Publishing Platform, and CMS"
 # wp import "${root}/env/data.xml" --authors=create
 
 wp option update show_on_front 'page'
-wp option update page_on_front 8891
+
+# Import content from WordPress.org
+php env/import-content.php --url 'https://wordpress.org/wp-json/wp/v2/posts?context=wporg_export&per_page=50'
+php env/import-content.php --url 'https://wordpress.org/wp-json/wp/v2/pages?context=wporg_export&per_page=50'
```
dd32: Instead of duplicating it, we can just call the refresh script.
```suggestion
# Import content from WordPress.org
./refresh.sh
```

dd32:
0b605144539d9dea7c059c8c3d131c7af28d3dd9 Well that was silly, that's my fault... that resulted in #630 :)

PR #623, opened 2026-03-04
Update app links

### LINKED ISSUE #622: Old app download links
The mobile and download page contains old links to the Apple App Store and to Google Play.
Apple links to iTunes and Google Play does not use https.

timse201:
fixes #622


dd32:
FYI @timse201 These patterns are actually generated automatically from the WordPress.org page source, by a github action, so a followup PR will likely change some of these now that I've updated the canonical source.

